Issue #31, opened 2017-10-18
ERROR: missing temporal_convolution_tbc, run `python setup.py install`


travel-go:
Hello,I am a green hand.when i use
CUDA_VISIBLE_DEVICES=6 python train.py data-bin/iwslt14.tokenized.de-en --lr 0.25 --clip-norm 0.1 --dropout 0.2 --max-tokens 4000 --arch fconv_iwslt_de_en --save-dir checkpoints/fconv

I meet a error .

ERROR: missing temporal_convolution_tbc, run `python setup.py install`
Traceback (most recent call last):
  File "train.py", line 14, in <module>
    from fairseq import bleu, data, options, utils
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/options.py", line 11, in <module>
    from fairseq import models
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/models/__init__.py", line 9, in <module>
    from . import fconv
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/models/fconv.py", line 14, in <module>
    from fairseq.modules import BeamableMM, LinearizedConvolution
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/modules/__init__.py", line 10, in <module>
    from .conv_tbc import ConvTBC
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/modules/conv_tbc.py", line 18, in <module>
    raise e
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/modules/conv_tbc.py", line 14, in <module>
    from fairseq import temporal_convolution_tbc
  File "/home/hfyu/fairseq16/fairseq-py-master/fairseq/temporal_convolution_tbc/__init__.py", line 3, in <module>
    from ._temporal_convolution_tbc import lib as _lib, ffi as _ffi
ImportError: /home/hfyu/fairseq16/fairseq-py-master/fairseq/temporal_convolution_tbc/_temporal_convolution_tbc.so: undefined symbol: _ZNK2at4Type4copyERKNS_6TensorERS1

I donnot know how to deal with it ,Could you help me?Thankyou.

myleott:
You need to run `python setup.py build` followed by `python setup.py develop`.

travel-go:
I use python setup.py build 

running install
running bdist_egg
running egg_info
writing fairseq.egg-info/PKG-INFO
writing dependency_links to fairseq.egg-info/dependency_links.txt
[...]

PR #42, opened 2017-10-24
Fix for building under clang: specify C++ build and use C++ linkage

jamesr66a:


myleott:
This should hopefully fix some of the issues brought up by @stovenator in #31 as well.

PR #46, opened 2017-10-30
Adding a note about docker flags.

tdomhan:
The note is taken from https://github.com/pytorch/pytorch#from-source. If someone uses fairseq with docker and does not set these flags he will get a very obscurve `Bus error`.


myleott:
Hi @tdomhan, thanks for submitting this! Unfortunately it seems we require the CLA (referenced above), even for a trivial change like this :/ If you want to sign the CLA, I will merge your PR. Otherwise, since it's a minor change, I can just do it myself.

tdomhan:
The latter sounds like the better solution ;) Feel free to close the PR and make the modification :)

PR #68, opened 2017-11-22
fix bug in lstm model

Zrachel:
The lstm model losses a projection from lstm hidden state to out_embed. Is it a bug?

### DIFF
--- a/fairseq/models/lstm.py
+++ b/fairseq/models/lstm.py
@@ -122,7 +122,8 @@ class LSTMDecoder(FairseqIncrementalDecoder):
             for layer in range(num_layers)
         ])
         self.attention = AttentionLayer(encoder_embed_dim, embed_dim)
-        self.fc2 = Linear(embed_dim, out_embed_dim)
+        if embed_dim != out_embed_dim:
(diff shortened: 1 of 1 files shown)


myleott:
Good catch! We were trying to follow the implementation here [1], which doesn’t have the extra projection. But as you noticed, the current fairseq version doesn’t work if out_embed_dim != embed_dim. How about if we conditionally apply the projection only when out_embed_dim != embed_dim?

[1] https://github.com/harvardnlp/seq2seq-attn/

fairseq/models/lstm.py
```diff
@@ -197,6 +199,9 @@ def _forward(self, input_tokens, encoder_out):
         attn_scores = attn_scores.transpose(0, 2)
 
         # project back to size of vocabulary
+        if hasattr(self, 'additional_fc'):
+            x = self.additional_fc(x)
+            x = F.dropout(x, p=self.dropout, training=self.training)
```
myleott: I think we want `p=self.dropout_out` here.

Zrachel: Sorry, I forgot I've changed to one dropout in my own code.

PR #75, opened 2017-12-02
Fixed 2 typos

toothlessdragon:
This is really cool work and I found these typos while reading through it.

To make sure there weren't any more, I also spellchecked it all with PyCharm.

Gotta start my open source career somewhere 🕴


myleott:
Good catch, thanks!

PR #77, opened 2017-12-06
Merge internal changes

### LINKED ISSUE #73: weight normalization dimension different in encoder and decoder convolutions
Is there any reason why the convolution filter weights are normalized differently in encoder vs decoder during training (I am referring to weight normalization being done across different dimensions in encoder and decoder). Does it make a difference in training if one normalizes the decoder conv filters the same way as in encoder (convTBC) ? My understanding from the original weight norm paper was that weight normalization is done per output channel, whereas this does not seem to be the case in decoder of FairSeq convolutional network.

myleott:
Changes:
- a233fce: Catch and recover from OOM in some cases
- 9f3ccaa: Fix weight norm dimension in decoder (fixes #73)
- 0a83627: Fix conv padding for even kernel widths
- Misc other features

### DIFF
--- a/fairseq/models/fairseq_decoder.py
+++ b/fairseq/models/fairseq_decoder.py
@@ -18,3 +18,6 @@ class FairseqDecoder(nn.Module):
     def max_positions(self):
         """Maximum input length supported by the decoder."""
         raise NotImplementedError
+
+    def upgrade_state_dict(self, state_dict):
+        return state_dict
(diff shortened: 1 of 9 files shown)


ngimel:
Cool fixes! While you are at it, it seems your softmaxes still don't have explicitly specified dimension, and new pytorch softmax interface (0.3 release and current master) deprecates softmax without explicitly specified dimension.

myleott:
We actually made that change previously but then reverted it -- it turns out old versions of PyTorch don't have the `dim` kwarg, so satisfying the warning would break backward compatibility.

At some point we need to just bump our version number of break compatibility with PyTorch <0.3.0, but may do that concurrently with #66.

Issue #88, opened 2018-01-03
Please provide preprocessed and binarized training set as well


dagarcia-nvidia:
We are trying to reproduce your results and it would be easier if you provided preprocessed and binarized training sets in addition to [the testing sets you have already made available](https://github.com/facebookresearch/fairseq-py#pre-trained-models).

Thanks!

edunov:
Hi @dagarcia-nvidia 

I'm working on a script that will download and prepare the data. Would it be helpful?

dagarcia-nvidia:
@edunov Sure, definitely. Thanks for the effort.

lukmaz:
Hi @edunov ,
Are you going to publish the scripts for the data preprocessing within the next 2 weeks?
Thanks,
Lukasz

lukmaz:
Thanks!

lukmaz:
Hi @edunov ,
I tried to test the scripts for training WMT14 en-de, and I think some paths are messed up: 
IWSLT14 is from DE to EN,
WMT14 is from EN to DE,
but both scripts produce files ...de-en...bin/idx, e.g:
data-bin/iwslt14.tokenized.de-en/train.de-en.de.bin
data-bin/wmt14_en_de/train.de-en.de.bin

lukmaz:
Hi @edunov ,
I played with the scripts for some time and I noticed, that WMT14_en_de script is missing test data. Are the same test data as for WMT14_en_fr going to work? I mean adding:
"http://statmt.org/wmt14/test-full.tgz" to URLS and
"test-full.tgz" to FILES

edunov:
Thanks @lukmaz I fixed it in the latest version. 
You're correct, both scripts share the same test file

edunov:
#91 is now landed, if there is anything else needed from us, please let us know
meanwhile, I'll close this issue

PR #91, opened 2018-01-12
Prepare scripts for WMT14 (#88)

edunov:
I'm testing these scripts at the moment. Once the testing is done, I'll update and land this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ The following command-line tools are available:
 ## Evaluating Pre-trained Models
 First, download a pre-trained model along with its vocabularies:
 ```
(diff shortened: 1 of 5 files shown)


bastings:
@edunov if you switch to News Commentary v12 then the results do not conform to the WMT14 translation task anymore and may not be comparable. Are you sure that's what you want to do?

edunov:
@bastings you're right, the reason to change it is to make our dataset consistent with others, in particular, the dataset used in "Attention is all you need" paper:  https://github.com/tensorflow/tensor2tensor/blob/master/tensor2tensor/data_generators/translate_ende.py#L42 

Personally, for future research purposes, I'd prefer to switch to the pre-processed data for WMT'18 as soon as it becomes available.

Issue #132, opened 2018-03-25
"index out of range" when preparing WMT14en2de


hitvoice:
When running 
```sh
cd data
bash prepare-wmt14en2de.sh
```
An "IndexError" is reported:
```
Tokenizer Version 1.1
Language: de
Number of threads: 8

splitting train and valid...
learn_bpe.py on wmt14_en_de/tmp/train.de-en...
Traceback (most recent call last):
  File "subword-nmt/learn_bpe.py", line 254, in <module>
    main(args.input, args.output, args.symbols, args.min_frequency, args.verbose, is_dict=args.dict_input)
  File "subword-nmt/learn_bpe.py", line 199, in main
    vocab = dict([(tuple(x[:-1])+(x[-1]+'</w>',) ,y) for (x,y) in vocab.items()])
  File "subword-nmt/learn_bpe.py", line 199, in <listcomp>
    vocab = dict([(tuple(x[:-1])+(x[-1]+'</w>',) ,y) for (x,y) in vocab.items()])
IndexError: string index out of range
apply_bpe.py to train.en...
apply_bpe.py to valid.en...
apply_bpe.py to test.en...
apply_bpe.py to train.de...
apply_bpe.py to valid.de...
apply_bpe.py to test.de...
clean-corpus.perl: processing wmt14_en_de/tmp/bpe.train.en & .de to wmt14_en_de/train, cutoff 1-250, ratio 1.5
..........(100000)..........(200000)..........(300000)..........(400000)..........(500000)..........(600000)..........(700000)..........(800000)..........(900000)..........(1000000)..........(1100000)..........(1200000)..........(1300000)..........(1400000)..........(1500000)..........(1600000)..........(1700000)..........(1800000)..........(1900000)..........(2000000)..........(2100000)..........(2200000)..........(2300000)..........(2400000)..........(2500000)..........(2600000)..........(2700000)..........(2800000)..........(2900000)..........(3000000)..........(3100000)..........(3200000)..........(3300000)..........(3400000)..........(3500000)..........(3600000)..........(3700000)..........(3800000)..........(3900000)..........(4000000)..........(4100000)..........(4200000)..........(4300000)..........(4400000)..........(4500000)....
Input sentences: 4544200  Output sentences:  3557596
clean-corpus.perl: processing wmt14_en_de/tmp/bpe.valid.en & .de to wmt14_en_de/valid, cutoff 1-250, ratio 1.5
....
Input sentences: 45901  Output sentences:  36073
[...]

PR #134, opened 2018-03-28
Update training commands

hitvoice:
Update training commands in data/README to match the latest version of this project according to #132.

- Motivation: in the previous data/README, the commands are obsolete and will cause the error "unrecognized arguments: --label-smoothing 0.1 --force-anneal 50". 
- What's changed: add arguments "--criterion label_smoothed_cross_entropy" and "--lr-scheduler fixed" to the training commands of all 3 datasets.
- Result: the new commands run without error on all 3 datasets.

### DIFF
--- a/data/README.md
+++ b/data/README.md
@@ -18,9 +18,9 @@ $ python preprocess.py --source-lang de --target-lang en \
   --trainpref $TEXT/train --validpref $TEXT/valid --testpref $TEXT/test \
   --destdir data-bin/iwslt14.tokenized.de-en
 
-# Train the model:
+# Train the model (better for a single GPU setup):
 $ mkdir -p checkpoints/fconv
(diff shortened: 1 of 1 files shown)


data/README.md
```diff
@@ -20,9 +20,10 @@ $ python preprocess.py --source-lang de --target-lang en \
 
 # Train the model:
 $ mkdir -p checkpoints/fconv
-$ CUDA_VISIBLE_DEVICES=0 python train.py data-bin/iwslt14.tokenized.de-en \
+$ python train.py data-bin/iwslt14.tokenized.de-en \
```
edunov: I'd put this back, the settings are only good for a single GPU setup, if someone with 8 GPUs runs this line, the resulting model performance will be less good.

hitvoice: I see. I put the flag back and add a line of explanation in the comment to make this intention clearer.

PR #142, opened 2018-04-17
Make sure setup.py is always run under its base directory

theweiho:
Regardless of where the script is called from


theweiho:
@myleott

myleott:
I think we can approach packaging a different way, I'll message you separately. Closing for now :)

PR #151, opened 2018-05-08
Add pretrained embedding support

sai-prasanna:
I have added new flags to set pre-trained embeddings for encoder and decoder.

Each line of the pre-trained embedding file should word followed by the dimension values of the corresponding vector.
Example for embedding with 4 dimensions.
```
Hi 0.2 0.1 0.7
This 0.33 0.22 0.44
```

### DIFF
--- a/fairseq/models/lstm.py
+++ b/fairseq/models/lstm.py
@@ -28,10 +28,14 @@ class LSTMModel(FairseqModel):
                             help='dropout probability')
         parser.add_argument('--encoder-embed-dim', type=int, metavar='N',
                             help='encoder embedding dimension')
+        parser.add_argument('--encoder-embed-path', default=None, type=str, metavar='STR',
+                            help='path to pre-trained encoder embedding')
         parser.add_argument('--encoder-layers', type=int, metavar='N',
                             help='number of encoder layers')
         parser.add_argument('--decoder-embed-dim', type=int, metavar='N',
                             help='decoder embedding dimension')
+        parser.add_argument('--decoder-embed-path', default=None, type=str, metavar='STR',
+			                help='path to pre-trained decoder embedding')
         parser.add_argument('--decoder-layers', type=int, metavar='N',
                             help='number of decoder layers')
         parser.add_argument('--decoder-out-embed-dim', type=int, metavar='N',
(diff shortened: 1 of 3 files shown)


fairseq/models/lstm.py
```diff
@@ -28,10 +28,14 @@ def add_args(parser):
...
         parser.add_argument('--encoder-embed-dim', type=int, metavar='N',
                             help='encoder embedding dimension')
+        parser.add_argument('--encoder-embed-path', default=None, type=str, metavar='STR',
+                            help='path to pre-trained encoder embedding')
         parser.add_argument('--encoder-layers', type=int, metavar='N',
                             help='number of encoder layers')
         parser.add_argument('--decoder-embed-dim', type=int, metavar='N',
                             help='decoder embedding dimension')
+        parser.add_argument('--decoder-embed-path', default=None, type=str, metavar='STR',
+			                help='path to pre-trained decoder embedding')
```
myleott: nit: alignment

fairseq/utils.py
```diff
@@ -248,6 +249,29 @@ def load_align_dict(replace_unk):
...
+            pieces = line.strip().split()
+            embed_dict[pieces[0]] = [float(weight) for weight in pieces[1:]]
+    return embed_dict
+
+def load_embedding(embed_dict, vocab, embedding):
+    embed_np = embedding.weight.data.numpy()
+    for idx in range(len(vocab)):
+        token = vocab[idx]
+        if token in embed_dict:
+            embed_np[idx] = np.array(embed_dict[token])
```
myleott: After the above change, we can just do: `embedding.weight.data[idx] = embed_dict[token]`. Then we can get rid of all the numpy usage.

fairseq/utils.py
```diff
@@ -248,6 +249,29 @@ def load_align_dict(replace_unk):
...
+def print_embed_overlap(embed_dict, vocab_dict):
+     embed_keys = set(embed_dict.keys())
+     vocab_keys = set(vocab_dict.symbols)
+     overlap = len(embed_keys & vocab_keys)
+     print("| Found {}/{} types in embedding file.".format(overlap, len(vocab_dict)))
+
+def parse_embedding(embed_path):
+    embed_dict = dict()
+    with open(embed_path) as f_embed:
+        for line in f_embed:
```
myleott: I think it'd be more generally useful to adhere to the fastText vector format, i.e., skip the header row. Something like (not tested):
```
with open(embed_path) as f_embed:
    header = next(f_embed)  # num_embed, embed_dim
    for line in f_embed:
        (...)
```

fairseq/utils.py
```diff
@@ -248,6 +249,29 @@ def load_align_dict(replace_unk):
...
+            pieces = line.strip().split()
+            embed_dict[pieces[0]] = [float(weight) for weight in pieces[1:]]
```
myleott: Let's return Tensors:
```
embed_dict[pieces[0]] = torch.Tensor([float(weight) for weight in pieces[1:]])
```

sai-prasanna:
@myleott  Thanks for the quick review. I made all the necessary changes.

PR #162, opened 2018-05-22
Add tqdm + no_progress_bar option for data loading progress bar

liezl200:


liezl200:
abandoning in favor of binarized data loading in PyTorch Translate

PR #171, opened 2018-05-31
Stop pretrained embeddings from loading during inference

sai-prasanna:
In interactive.py and generate.py , if we try to load a model which was trained pre trained embeddings, it will try to load the files, eventhough the model checkpoints files we give have the embedding weights.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -43,7 +43,10 @@ def main(args):
 
     # Load ensemble
     print('| loading model(s) from {}'.format(', '.join(args.path)))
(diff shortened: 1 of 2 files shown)


myleott:
We have a big update coming soon, where we've reworked pretrained embeddings (among many other changes), so I'll incorporate a fix there but it'll need to be slightly more general, since not all models may support pretrained embeddings.

I'll keep this open for now and will close it once the update lands.

myleott:
Just following up on this. You can now accomplish this with `--model-overrides`, for example:
```--model-overrides "{'encoder_embed_path':None}"```

PR #177, opened 2018-06-13
Implement multithread preprocess

paramecinm:
Add flag: workers. Use multiprocess to parallelize Tokenizer.binarize.

### DIFF
--- a/fairseq/tokenizer.py
+++ b/fairseq/tokenizer.py
@@ -7,6 +7,9 @@
 
 from collections import Counter
 import re
+import multiprocessing
+import pickle
+import tempfile
 
 import torch
 
@@ -40,8 +43,16 @@ class Tokenizer:
                 dict.add_symbol(dict.eos_word)
 
     @staticmethod
-    def binarize(filename, dict, consumer, tokenize=tokenize_line,
+    def binarize(filename, dict, consumer, worker_cnt=1, tokenize=tokenize_line,
                  append_eos=True, reverse_order=False):
+        if worker_cnt == 1:
+            return Tokenizer.binarize_sequential(filename, dict, consumer, tokenize, append_eos, reverse_order)
+        else:
+            return Tokenizer.binarize_parallel(filename, dict, consumer, worker_cnt, tokenize, append_eos, reverse_order)
+
+    @staticmethod
+    def binarize_sequential(filename, dict, consumer, tokenize=tokenize_line,
+            append_eos=True, reverse_order=False):
         nseq, ntok = 0, 0
         replaced = Counter()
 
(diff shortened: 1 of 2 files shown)


huihuifan:
Can you provide some numerical measurements of timing for binarization of wmt en-de on 1 worker, etc?

paramecinm:
On my macbook with 1 core, 4 processors.

worker count | run time(s) | run time with out build dictionary | speed up | speed up (without build dictionary)
-- | -- | -- | -- | --
1 | 1734 | 1522 | 1 | 1
2 | 1235 | 1023 | 1.40405 | 1.48778
4 | 909 | 697 | 1.90759 | 2.18364
8 | 815 | 603 | 2.12761 | 2.52405

I am running experiments on dev server with more workers, and try to figure out if it is necessary to parallelize building dictionary.

edunov:
As discussed offline, let's test performance on more cores say 20-40. And let's check if we need to parallelize dictionary construction.

paramecinm:
I have parallelized dictionary build, and here is the new speed up on fb dev server, who have 24 cores.

worker_cnt | time | speedup
-- | -- | --
1 | 2263 | 1
12 | 591 | 3.8291
16 | 550 | 4.11455
24 | 517 | 4.37718
32 | 527 | 4.29412
48 | 532 | 4.25376

bouvrie:
Thanks for posting this PR; it'd be a great addition! It looks like in the proposed approach all the workers  simultaneously attempt to scan the entire input file. You might also consider benchmarking an approach where the parent process (and only the parent) reads the file and deals groups of lines to the workers through shared memory.

paramecinm:
@bouvrie Python's shared memory can only share fix length array with basic types. To share strings is not supported by default. Implementing producer-consumer model will also have overhead. Also I have recored the time for each part of the program, on WMT dataset, reading the whole file will only take very few seconds while binarizing and merging worker result will take hundreds of seconds. So I think it won't be worthy to optimize the file reading part.

bouvrie:
@paramecinm Thanks, fair point on the file loading. I took a closer look at `Tokenizer.tokenize` and with the default `tokenize_line` it looks like most of the time is possibly spent in everything except the tokenizing (dict lookups, publishing to the `replaced` consumer). Perhaps this is the source of the contention/bottleneck in the many-thread setting?

paramecinm:
@bouvrie Because the python shared memory in multiprocess doesn't support share complex structures, and using Python's data sharing api like Pipe to pass data is even slower than write data to file and read them in. So I choose to write down the partial result of each process and read them to merge in main process. The write and merge part takes a large amount of time. To further optimize the performance, I can only think of use the much powerful multithread in C. It will be great if you can find some better approach!

edunov:
I wonder if we will have a better performance if master instead of waiting for workers to finish binarization will consume their output as they produce it. It can be done through piping. Right now, pickle.load() seems to kill a lot of performance. Also, it should save memory, as you won't need to keep everything in ram.

myleott:
This is implemented via the `--workers` option to preprocess.py

PR #180, opened 2018-06-19
add count for padding words

ngimel:
Otherwise, even with padding_factor specified, dictionary size remains non-multiple of padding factor.

### DIFF
--- a/fairseq/data/dictionary.py
+++ b/fairseq/data/dictionary.py
@@ -125,6 +125,7 @@ class Dictionary(object):
             i = 0
             while threshold_nwords % padding_factor != 0:
                 new_symbols.append('madeupword{:04d}'.format(i))
+                new_count.append(0)
                 i += 1
                 threshold_nwords += 1
 


myleott:
This just affects the dictionary file that's output, right? Internally the dictionary is still a multiple of the padding factor..

ngimel:
I didn't dig deep, but when dict.en.txt and dict.de.txt weren't multiples of 8, the actual training saw non-multiple-of 8 dict size (train.py does not accept padding_factor argument) and called non-hmma gemms. May be I was doing something wrong?

myleott:
Ah, you're right. We generally preprocessed our datasets so that they'd be multiples of 8, so didn't notice that `padding_factor` didn't quite work.

fairseq/data/dictionary.py
```diff
@@ -125,6 +125,7 @@ def finalize(self, threshold=1, nwords=-1, padding_factor=8):
             i = 0
             while threshold_nwords % padding_factor != 0:
                 new_symbols.append('madeupword{:04d}'.format(i))
+                new_count.append(1)
```
myleott: Let's make this 0 instead of 1, to make it clearer that these words don't actually appear in the dataset.

PR #190, opened 2018-06-21
Fix `--output-format raw` option to preprocess.py (Fixes #188)

### LINKED ISSUE #188: the raw output data can not be trained
when I use preprocess.py to ouput raw data by use `--output-format raw`, the output data file can not be used to train.py which I also use `--raw-text`. Through looking the source code, I change the output data file mentioned above in this way: rename 'train.src' to 'train.src-tgt.src' and  'train.tgt' to 'train.src-tgt.tgt' (assume I use `--source-lang=src --target-lang=tgt` ) and this can run.
I think it's a bug and is easy to fix :)

myleott:


michaelauli:
what do we use this flag for?

myleott:
If you preprocess a dataset with `--output-format raw` then the resulting dataset will be plaintext (i.e., not binarized). You can then train a model with this data by passing the `--raw-text` option to train.py

PR #193, opened 2018-06-21
Minor fix

hmc-cs-mdrissi:
The current version of eval_lm, while it accepts a task only really works if the task language_modeling since tokens_per_sample isn't an option for translation. This one line fix makes it work for translation.


alexeib:
looks fine to me, thanks

PR #211, opened 2018-07-05
Default model options

hmc-cs-mdrissi:
This deals with issue 181. Generally I removed defaults from parser.add_argument and instead added those defaults to where the model architectures were registered. The main exception is for the fconv_self_att_wp model where I made it's defaults consistent with what is in the stories readme so that'd it match closer with the paper. I didn't change the readme, although if wanted I can simplify the train line a bit now as the model options used there now consider to the ones it'll have without specifying them.


hmc-cs-mdrissi:
@huihuifan, One discrepancy I've noticed and am not sure what's the source of it is the number of parameters for the wp model is about 100 million while trying to match hyperparameter values with the paper. The paper has about another 34 million parameters.

Issue #213, opened 2018-07-10
Size Mismatch in AdaptiveSoftmax when targets are not specified


amrsharaf:
Following up on #212 , I'm updating `sequence_generator.py` to generate text from a pre-trained language model (initially trained with adaptive softmax). When computing log probabilities, and the targets are set to none, I receive a size mismatch exception in the line below, possibly because the dictionary size is smaller than the adaptive softmax cut-off: 

https://github.com/pytorch/fairseq/blob/388c520be21752cacb9fe3b1712038f32e0e9a5f/fairseq/modules/adaptive_softmax.py#L126

I imagine this could be solved by some sort of truncation to the output of tail[i].input

huihuifan:
Why is the dictionary size smaller than the cutoff for adaptive softmax? Isn't the adaptive softmax cutoff's last bin value the size of the dictionary?

amrsharaf:
I'm checking now, but I don't this this is the case, the last cut-off is an upper bound on the vocabulary size: https://github.com/pytorch/fairseq/blob/388c520be21752cacb9fe3b1712038f32e0e9a5f/fairseq/modules/adaptive_softmax.py#L23

alexeib:
@amrsharaf how come your dictionary is smaller than the adaptive softmax? its true that if you specify adaptive softmax cutoff of 500,000 and your vocab size is 5,000 then you will get an adaptive softmax of size 500k, where only the first 5k is useful. you can work around that by taking the first ```vocab_size``` columns from what is returned by adaptive softmax, but in general the idea was that you should create adaptive softmax cutoffs that are compatible with your vocab size. we can put some assert to make sure that vocab size is at least as big as the last specified cutoff

amrsharaf:
That's true, I should have made the max cut-off == vocab size, I just used the same hyperparameters from the wikitext README tutorial, it'd be helpful to add the assert statement or just an if statement like this: 

`if vocab_size < cutoff[i]: cutoff[i] = vocab_size`

Issue #212, opened 2018-07-09
Assertion failure when generating text from LM with Adaptive softmax


amrsharaf:
I'm updating the `generate.py` to generate text from a language model pre-trained with an adaptive softmax function. Generation crashes with an assertion failure in the line below:

https://github.com/pytorch/fairseq/blob/1d79ed9b5f67a51e468d012b2ba1bded6d8ebbaf/fairseq/models/fconv.py#L497

I checked the adaptive softmax code and it's seems the targets are only needed for the forward path, is it safe to ignore the assertion and set the `sample[target]` to `None` in the `get_normalized_probs` function?

alexeib:
targets are needed during training and when computing perplexities to make it faster. When you provide targets, you will only get probabilities back for the targets themselves and not anything else.

for generation you want the probability distribution over all toens (presumably). the easiest way to get that from adaptive softmax is, as you suggest, to have target in the sample set to 'None' which tells adaptive softmax that there are no specific targets. 

also if you get generation implemented and working, pull requests are welcome :)

amrsharaf:
Thank you so much. I'll submit a pull request as soon as I finish this.

PR #215, opened 2018-07-11
update sequence generator to support generation from a language model

amrsharaf:
This is a simple update for the `sequence_generator.py` script to support generation from a language model trained on the target vocabulary. Changes include:

1) Clone `max_positions` for source / target if it's set by a monolingual  LM;
2) Set the encoder output dictionary to None when using LM;
3) Check for None values of encoder output dictionary when calling `reorder_encoder_out`
4) Assert targets are given for adaptive softmax only at training time;
5) Copy attention values only if provided by the decoder model;
6) Squeeze the `probs` tensor instead of `decoder_out` as the adaptive softmax probability function expects 3D inputs.

This is a follow up on the issues: #213 and #212. I'm basically replicating the LM results from the hierarchical story generation paper (https://arxiv.org/abs/1805.04833), but I can't find the LM generation code in the repo. I'd appreciate if @huihuifan can take a look at this, in case it's a replicated or redundant pull request.

### DIFF
--- a/fairseq/data/language_pair_dataset.py
+++ b/fairseq/data/language_pair_dataset.py
@@ -139,6 +139,10 @@ class LanguagePairDataset(FairseqDataset):
     def _get_max_positions(self, max_positions):
         if max_positions is None:
             return self.max_source_positions, self.max_target_positions
+        # If max_position is not a sequence (possibly from a monolingual model)
+        # then use the same upper bound for source & target
+        if not hasattr(max_positions, '__len__'):
+            max_positions = (max_positions, max_positions)
         assert len(max_positions) == 2
         max_src_pos, max_tgt_pos = max_positions
         return min(self.max_source_positions, max_src_pos), min(self.max_target_positions, max_tgt_pos)
--- a/fairseq/models/fconv.py
+++ b/fairseq/models/fconv.py
@@ -269,7 +269,7 @@ class FConvEncoder(FairseqEncoder):
         }
 
     def reorder_encoder_out(self, encoder_out_dict, new_order):
-        if encoder_out_dict['encoder_out'] is not None:
+        if encoder_out_dict is not None and encoder_out_dict['encoder_out'] is not None:
             encoder_out_dict['encoder_out'] = (
                 encoder_out_dict['encoder_out'][0].index_select(0, new_order),
                 encoder_out_dict['encoder_out'][1].index_select(0, new_order),
@@ -494,8 +494,13 @@ class FConvDecoder(FairseqIncrementalDecoder):
         """Get normalized probabilities (or log probs) from a net's output."""
 
         if self.adaptive_softmax is not None:
-            assert sample is not None and 'target' in sample
-            out = self.adaptive_softmax.get_log_prob(net_output[0], sample['target'])
+            if sample is not None:
+                # Assert enforced only at training time
+                assert 'target' in sample
+                out = self.adaptive_softmax.get_log_prob(net_output[0], sample['target'])
+            else:
+                # No target was provided, setting to None
+                out = self.adaptive_softmax.get_log_prob(net_output[0], target=None)
             return out.exp_() if not log_probs else out
         else:
             return super().get_normalized_probs(net_output, log_probs, sample)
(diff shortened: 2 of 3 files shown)


huihuifan:
The experiments from the story paper were actually run with luatorch, but I wanted a PyTorch code release because most people use PyTorch now. @alexeib implemented the convlm model in PyTorch as well (I had it in Lua as well when we originally worked on that paper). Will take a look right now. @alexeib as well?

fairseq/data/language_pair_dataset.py
```diff
@@ -139,6 +139,10 @@ def valid_size(self, index, max_positions):
     def _get_max_positions(self, max_positions):
         if max_positions is None:
             return self.max_source_positions, self.max_target_positions
+        # If max_position is not a sequence (possibly from a monolingual model)
+        # then use the same upper bound for source & target
+        if not hasattr(max_positions, '__len__'):
```
alexeib: can we check if its not type(tuple) instead ?

fairseq/models/fconv.py
```diff
@@ -494,8 +494,13 @@ def get_normalized_probs(self, net_output, log_probs, sample):
...
         if self.adaptive_softmax is not None:
-            assert sample is not None and 'target' in sample
-            out = self.adaptive_softmax.get_log_prob(net_output[0], sample['target'])
+            if sample is not None:
+                # Assert enforced only at training time
+                assert 'target' in sample
+                out = self.adaptive_softmax.get_log_prob(net_output[0], sample['target'])
+            else:
+                # No target was provided, setting to None
+                out = self.adaptive_softmax.get_log_prob(net_output[0], target=None)
```
alexeib: can we move "target" out into a variable and then we can just call get_log_prob once unconditionally? (to make refactors easier in the future)

fairseq/sequence_generator.py
```diff
@@ -268,7 +272,7 @@ def get_hypo():
                 for i, model in enumerate(self.models):
                     if isinstance(model.decoder, FairseqIncrementalDecoder):
                         model.decoder.reorder_incremental_state(incremental_states[model], reorder_state)
-                    encoder_outs[i] = model.encoder.reorder_encoder_out(encoder_outs[i], reorder_state)
+                    encoder_outs[i] = model.decoder.reorder_encoder_out(encoder_outs[i], reorder_state)
```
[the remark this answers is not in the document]
alexeib: i thought they are all on encoder now, which maes sense.

in case of a language model encoder_outs should be empty anyway, maybe we can just check that? and this particular change would probably cause all the encoder-decoder models to crash (e.g. machine translation)

fairseq/sequence_generator.py
```diff
@@ -521,9 +526,10 @@ def _decode_one(self, tokens, model, encoder_out, incremental_states, log_probs)
                 decoder_out = list(model.decoder(tokens, encoder_out, incremental_states[model]))
             else:
                 decoder_out = list(model.decoder(tokens, encoder_out))
-            decoder_out[0] = decoder_out[0][:, -1, :]
             attn = decoder_out[1]
             if attn is not None:
                 attn = attn[:, -1, :]
         probs = model.get_normalized_probs(decoder_out, log_probs=log_probs)
+        probs = probs[:, -1, :]
```
alexeib: why move it here? its probably much faster to get probs just for the last state than for everything, and then take the last one?

fairseq/sequence_generator.py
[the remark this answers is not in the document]
huihuifan: oops yes I forgot we agreed to move them all into the encoder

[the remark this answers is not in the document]
amrsharaf: Yes, this should be `encoder` instead of `decoder`, I'll fix it.

@huihuifan , @alexeib Thank you so much for looking into this. I'll address all the comments and run test to make sure the changes doesn't break generation for translation.

fairseq/data/language_pair_dataset.py
alexeib: sorry i meant check if type is int here

amrsharaf: max_positions was actually a float, I think this check is more general as it capture both tuple and list...

fairseq/sequence_generator.py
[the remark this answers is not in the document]
huihuifan: thanks!

myleott:
Cool! @alexeib already covered all the comments I'd have. Please also make sure the integration tests pass: `python setup.py test`.

PR #230, opened 2018-07-26
Correct path in the pre-processing example

alvations:
Initially it's pointing to some arbitrary `/data/iwslt14.tokenized.de-en`, but given the steps above to cd `examples/translation` and then `cd ../..`, the appropriate path that contains the `iwslt14.tokenized.de-en` is `examples/translation/iwslt14.tokenized.de-en`


myleott:
Thanks!

PR #234, opened 2018-07-31
Correct the help name of the prefixes arguments

alvations:
The helper messages / description of the arguments are not exactly right for `preprocess.py`:

**Currently**: 

```python
parser.add_argument('--trainpref', metavar='FP', default=None, help='target language')
parser.add_argument('--validpref', metavar='FP', default=None, help='comma separated, valid language prefixes')
parser.add_argument('--testpref', metavar='FP', default=None, help='comma separated, test language prefixes')
```

In details:

 - `--trainpref` isn't the target language, that's the `-t` / `--target-language` argument
     - It should be the training file prefix, hence `help='train file prefix'`  edit
 - `--validpref` and `--testpref` isn't really the language prefix, it's the file prefix before the language, hence:
     - `help='comman separated, valid file prefixes'`
     - `help='comman separated, test file prefixes'`


myleott:
Good call, thanks!

Issue #223, opened 2018-07-20
How to ensemble Fairseq models?


MaggieMeow:
Parts of the code suggests that generate.py can generate sequences based on multiple models. 

https://github.com/pytorch/fairseq/blob/30ef667d40b9b34ca9caa43706a89b57531f19a7/generate.py#L41

Could you please kindly suggest an example of this usage, if possible?

Thanks!

myleott:
Yes, if you pass `--path model1.pt:model2.pt` it will ensemble model1 and model2 for generation. Note that ensembling this way loads both models into memory, so it may be necessary to do ensemble generation on CPU with the `--cpu` flag (e.g., if the models are too big to store multiple of them in GPU memory).

MaggieMeow:
I see. Thanks a lot!

MaggieMeow:
@myleott I couldn't ensemble two models from two different architectures, e.g. a LSTM and a transformer. It trows a runtime error: Error(s) in loading state_dict for LSTMModel. Any workarounds to get the ensemble to work for models from 2 different architectures?

Thanks!

huihuifan:
closing, merged into master. thanks!

PR #235, opened 2018-07-31
Add ensemble for different architectures

MaggieMeow:
Resolves the problem described in #223

### DIFF
--- a/fairseq/utils.py
+++ b/fairseq/utils.py
@@ -146,13 +146,16 @@ def load_ensemble_for_inference(filenames, task, model_arg_overrides=None):
         state = torch.load(filename, map_location=lambda s, l: default_restore_location(s, 'cpu'))
         state = _upgrade_state_dict(state)
         states.append(state)
-    args = states[0]['args']
+    args_list = []
+    for state in states:
+        args_list.append(state['args'])
     if model_arg_overrides is not None:
-        args = _override_model_args(args, model_arg_overrides)
+        for args in args_list:
+            args = _override_model_args(args, model_arg_overrides)
 
     # build ensemble
     ensemble = []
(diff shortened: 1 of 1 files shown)


fairseq/utils.py
```diff
@@ -146,13 +146,16 @@ def load_ensemble_for_inference(filenames, task, model_arg_overrides=None):
...
         state = _upgrade_state_dict(state)
         states.append(state)
-    args = states[0]['args']
+    args_list = []
+    for state in states:
+        args_list.append(state['args'])
     if model_arg_overrides is not None:
-        args = _override_model_args(args, model_arg_overrides)
+        for args in args_list:
+            args = _override_model_args(args, model_arg_overrides)
```
myleott: This doesn't quite work. It will reassign the locally-scoped args to the overridden version, but this will not be reflected in args_list (except that args is passed into _override_model_args by reference, so it might get changed in-place).

But anyway, let's get rid of args_list and this entire block above. We can do it inline below instead.

fairseq/utils.py
```diff
@@ -146,13 +146,16 @@ def load_ensemble_for_inference(filenames, task, model_arg_overrides=None):
...
 
     # build ensemble
     ensemble = []
-    for state in states:
+    for args, state in zip(args_list, states):
```
myleott: Replace with:
```
for state in states:
    args = state['args']
    if model_arg_overrides is not None:
        args = _override_model_args(state['args'], model_arg_overrides)
    model = task.build_model(args)
    (...)
```

MaggieMeow: Great idea. I will make the changes accordingly. Thanks!

myleott:
Thanks!

PR #240, opened 2018-08-08
add label smoothing for adaptive loss

haowei01:
Previously there is label smoothing for cross-entropy.  We want to add label smoothing for adaptive_loss.

1. refactored the adaptive_loss
2. added a LabelSmoothedAdaptiveLoss criterion.
3. add unit test

### DIFF
--- a/fairseq/criterions/adaptive_loss.py
+++ b/fairseq/criterions/adaptive_loss.py
@@ -32,13 +32,26 @@ class AdaptiveLoss(FairseqCriterion):
         """
 
         assert hasattr(model.decoder, 'adaptive_softmax') and model.decoder.adaptive_softmax is not None
-        adaptive_softmax = model.decoder.adaptive_softmax
 
         net_output = model(**sample['net_input'])
+
+        loss = self.compute_cross_entropy_loss(model, net_output, sample, reduce)
+
+        sample_size = sample['target'].size(0) if self.args.sentence_avg else sample['ntokens']
+        logging_output = {
+            'loss': utils.item(loss.data) if reduce else loss.data,
+            'ntokens': sample['ntokens'],
+            'sample_size': sample_size,
+        }
+        return loss, sample_size, logging_output
+
+    def compute_cross_entropy_loss(self, model, net_output, sample, reduce=True):
+        """Compute the cross_entropy loss from the net_output."""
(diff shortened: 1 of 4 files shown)


fairseq/criterions/adaptive_loss.py
```diff
@@ -32,13 +32,26 @@ def forward(self, model, sample, reduce=True):
...
+
+        sample_size = sample['target'].size(0) if self.args.sentence_avg else sample['ntokens']
+        logging_output = {
+            'loss': utils.item(loss.data) if reduce else loss.data,
+            'ntokens': sample['ntokens'],
+            'sample_size': sample_size,
+        }
+        return loss, sample_size, logging_output
+
+    def compute_cross_entropy_loss(self, model, net_output, sample, reduce=True):
```
alexeib: maybe instead of passing in model, we can pass in adaptive_softmax and target?

fairseq/criterions/label_smoothed_adaptive_loss.py
```diff
@@ -0,0 +1,61 @@
...
+        logging_output = {
+            'loss': utils.item(loss.data) if reduce else loss.data,
+            'cross_entropy': utils.item(cross_entropy.data) if reduce else cross_entropy.data,
+            'ntokens': sample['ntokens'],
+            'sample_size': sample_size,
+        }
+        return loss, sample_size, logging_output
+
+    def compute_smooth_loss(self, model, net_output, sample, reduce=True):
+        "Helper function to compute the smoothed loss"
```
alexeib: triple quotes?

fairseq/criterions/adaptive_loss.py
haowei01: it will become more complicated if we pass in adaptive_softmax and target, e.g. in the `forward` function of `AdaptiveLoss` and `LabelSmoothedAdaptiveLoss`, we will need to extract the `adaptive_softmax`, extract `target`, which makes the logic more duplicated.

In addition, in this way, the input args for `compute_smooth_loss ` is the same.

haowei01:
I am running on GPU, but the machine get restarted frequently.  It is probably easier to have this criterion implemented, and then use the workflow to test the model.

PR #244, opened 2018-08-09
add end-of-stack normalizations in case normalize_before has been set

ngimel:
tf's implementation with normalizations preceding ops also has normalizations at the end of stack. 
https://github.com/tensorflow/models/blob/master/official/transformer/model/transformer.py#L343
https://github.com/tensorflow/models/blob/master/official/transformer/model/transformer.py#L417
This PR makes fairseq consistent with tf, otherwise t2t config explodes.

### DIFF
--- a/fairseq/models/transformer.py
+++ b/fairseq/models/transformer.py
@@ -200,6 +200,7 @@ class TransformerEncoder(FairseqEncoder):
             TransformerEncoderLayer(args)
             for i in range(args.encoder_layers)
         ])
+        self.register_buffer('version', torch.Tensor([2]))
         self.normalize = args.encoder_normalize_before
         if self.normalize:
            self.layer_norm = LayerNorm(embed_dim)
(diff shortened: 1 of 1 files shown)


myleott:
Ah, good catch. I based this on the tensor2tensor implementation, which I also has the behavior you suggest: https://github.com/tensorflow/tensor2tensor/blob/d4040ae13e90e0a4e43c0669d6ce33ecd23cc0eb/tensor2tensor/models/transformer.py#L1241-L1244

Since some old model checkpoints may still depend on the old (incorrect) version, we don't want to break those checkpoints. Can you please:
1) Add a version buffer, similar to this: https://github.com/pytorch/fairseq/blob/f7f2dd01ea141cc3303d6735e0fdcf8402b1557b/fairseq/models/fconv.py#L351
2) Add an upgrade_state_dict method that checks the version and turns off the final normalization for old versions: https://github.com/pytorch/fairseq/blob/f7f2dd01ea141cc3303d6735e0fdcf8402b1557b/fairseq/models/fconv.py#L507-L515

myleott:
Thanks!

Issue #217, opened 2018-07-16
oom occurs when using too many training pairs?


zherowolf:
Hi, @myleott  Thanks for your great work on transformer with pytorch.
I have encounter with a oom problems when I try to train a translation model with more training pairs , like tens of millions or more.
I think the problem occurs because the training pairs will be read into memory at first. 
Did you train a model with ~100 millions pairs of sentences ?
It's OK with ~10 millions pairs of sentences.

By the way , I run with 4 V100 GPUs and the memory is 64G. the total training pairs ( binaried with preprocess.py) are about 15G.
but while loading datasets, the memory is almost fully occupied.

zherowolf:
I can run with changing class IndexedDataset to IndexedInMemoryDataSet when loading dataset splits ,But the training goes much slower.
Did you have some other advice ?
Thx.

zherowolf:
I tried using IndexedDataset and setting num_workers > 1, errors occured showing:
daemonic processes are not allowed to have children.
I will try to preprocess chunks of the data and report later.

edunov:
I'm working on a permanent solution meanwhile, you can use IndexedDataset instead of IndexedInMemoryDataset.

SinghJasdeep:
Thanks edunov! 

Where would I make this change? 

It seems like the default is IndexedDataset in preprocess.py?

(line 118) 
ds = indexed_dataset.IndexedDatasetBuilder(dataset_dest_path(output_prefix, lang, 'bin'))

myleott:
I think you would change this to IndexedDataset: https://github.com/pytorch/fairseq/blob/6381cc977fa909bd53c8915a18684e49b3507332/fairseq/tasks/translation.py#L117. It shouldn't be necessary to change preprocess.py

myleott:
You should be able to use `--fix-batches-to-gpus` (introduced in 90f52a1) to fix this.

PR #291, opened 2018-09-25
Merge internal changes

myleott:
Changelog:
- 4908863: Switch to DistributedDataParallelC10d and bump version 0.5.0 -> 0.6.0
  - no more FP16Trainer, we just have an FP16Optimizer wrapper
  - most of the distributed code is moved to a new wrapper class called DistributedFairseqModel, which behaves like DistributedDataParallel and a FairseqModel at the same time
  - Trainer now requires an extra dummy_batch argument at initialization, which we do fwd/bwd on when there's an uneven number of batches per worker. We hide the gradients from these dummy batches by multiplying the loss by 0
  - Trainer.train_step now takes a list of samples, which will allow cleaner --update-freq
- 1c56b58: parallelize preprocessing
- Misc bug fixes and features


SinghJasdeep:
Hey Myle does this include a solution for issue #217?  
@myleott 
https://github.com/pytorch/fairseq/issues/217

Thanks!

myleott:
Not yet, we're still iterating on that fix, since depending on how we implement it, it could reduce randomness and give worse results. We should have it ready soon though.

PR #293, opened 2018-09-27
Update sinusoidal_positional_embedding.py

zherowolf:
fixing bugs with error "argument must be tuple of ints, not Tensor"

### DIFF
--- a/fairseq/modules/sinusoidal_positional_embedding.py
+++ b/fairseq/modules/sinusoidal_positional_embedding.py
@@ -58,7 +58,7 @@ class SinusoidalPositionalEmbedding(nn.Module):
 
     def forward(self, input, incremental_state=None, timestep=None):
         """Input is expected to be of size [bsz x seqlen]."""
-        bsz, seq_len = torch.onnx.operators.shape_as_tensor(input)
+        bsz, seq_len = input.size()
         max_pos = self.padding_idx + 1 + seq_len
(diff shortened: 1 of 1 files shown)


jhcross:
@zherowolf, can you clarify which use of bsz or seq_len is causing the error?

jhcross:
We can't merge this change as-is because it breaks ONNX export to Caffe2 (see, e.g., https://github.com/pytorch/translate/blob/0728a48757436015b04ab9c3073dc5d65819a9a2/pytorch_translate/test/test_onnx.py#L216), but I'll be happy to fix the error in another way if you can provide details of the line where it is occurring.

zherowolf:
the bug occurs in [https://github.com/pytorch/fairseq/blob/master/fairseq/modules/sinusoidal_positional_embedding.py#L85](url)

zherowolf:
the bug occurs in https://github.com/pytorch/fairseq/blob/master/fairseq/modules/sinusoidal_positional_embedding.py#L85

jhcross:
Thanks @zherowolf ! And does it fix that error if instead of passing bsz and seq_len directly, you pass them through an `int` constructor:

`        return self.weights.index_select(0, positions.view(-1)).view(int(bsz), int(seq_len), -1).detach()
`

zherowolf:
great resolution. That works fine.

myleott:
Closing, since this seems resolved, thanks!

PR #304, opened 2018-10-04
multihead_attention: pre-transpose incremental state

jhcross:
Summary:
Though transpose operations are essentially free during PyTorch execution, they can result in costly operations when exported to Caffe2 inference nets via ONNX tracing, especially when applied repeatedly to large tensors.

For this reason, we update `MultiheadAttention` to store its incremental state with shape (bsz, num_heads, seq_len, head_dim), that is after transposing the projected input. This should result in non-trivially faster exported models without changing the semantics or speed of PyTorch execution.

Differential Revision: D10186506


myleott:
Landed as 265f42b7d52df824364f6709678a0827ef2f19e7

Issue #314, opened 2018-10-17
More Pythonic nn.Module object overloading


alvations:
The way Fairseq creates a function to overload the `nn.Embedding` object is a little awkward.  https://github.com/pytorch/fairseq/blob/master/fairseq/models/lstm.py#L448

```python
def Embedding(num_embeddings, embedding_dim, padding_idx):
    m = nn.Embedding(num_embeddings, embedding_dim, padding_idx=padding_idx)
    nn.init.uniform_(m.weight, -0.1, 0.1)
    nn.init.constant_(m.weight[padding_idx], 0)
    return m
```

Perhaps, this is better:

```python
class Embedding(nn.Embedding):
    def __init__(self, num_embeddings, embedding_dim, padding_idx):
        super().__init__(num_embeddings, embedding_dim, padding_idx=padding_idx)
        nn.init.uniform_(self.weight, -0.1, 0.1)
        nn.init.constant_(self.weight[padding_idx], 0)
```

Details on https://discuss.pytorch.org/t/why-does-code-like-fairseq-override-the-default-initialization-of-the-nn-embedding-layer/27358

alvations:
And for `nn.Linear` layer:

```python
class Linear(nn.Linear):
    def __init__(self, in_features, out_features, bias=True, dropout=0):
        super().__init__(in_features, out_features, bias=bias, dropout=dropout)
        self.weight.data.uniform_(-0.1, 0.1)
        if bias:
            self.bias.data.uniform_(-0.1, 0.1)
```

alvations:
And for `nn.Linear`:

```python
class Linear(nn.Linear):
    def __init__(self, in_features, out_features, bias=True):
        super().__init__(in_features, out_features, bias=bias)
        self.weight.data.uniform_(-0.1, 0.1)
        if bias:
            self.bias.data.uniform_(-0.1, 0.1)
```

Oddly, currently code at https://github.com/pytorch/fairseq/blob/master/fairseq/models/lstm.py#L471
has an unused `dropout=0` in the `**kwargs`.

myleott:
Hey, thanks for bring this up! We ended up doing a similar (slightly larger) refactoring and will merge that shortly. Closing this for now.

PR #315, opened 2018-10-17
More pythonic nn.Module overloading for fairseq.models.lstm

alvations:
Patching #314


myleott:
Will merge a slightly larger refactor along these lines shortly.

PR #339, opened 2018-11-01
Fix "ignore-case" behavior

ngimel:
Currently, if `ignore-case` is set, the same line will be yielded twice - once as lower-cased version, once as original version, leading to lower than expected uncased scores.


myleott:
Wow, nice catch! I think we never use this feature, but it looks like it's been wrong since the beginning.

PR #364, opened 2018-11-14
Myserver

beichao1314:
add Translation class to make server.


beichao1314:
Sorry, I make a mistake.

PR #373, opened 2018-11-17
Do LayerNorm in FP32 (fixes #335)

### LINKED ISSUE #335: layer normalization fp16 big transformer
I've been going through the "Scaling Neural Machine Translation" paper, and according to the paper the entire forward pass happens in fp16, not including the loss calculation.  However it does include the layer norm calculation.  I'm wondering, does this not cause any reduction overflows?  My understanding of fp16 training is that ops that do large reductions such as layer norm, batch norm, softmax, etc, should be done in fp32 for safety to avoid these overflows.  Is there something else going on here to handle this when training the big transformer?

myleott:
See discussion in #335.

I retrained WMT'16 En-De and get nearly identical results. This approach is safer but is also a bit slower.

### DIFF
--- a/fairseq/models/transformer.py
+++ b/fairseq/models/transformer.py
@@ -754,9 +754,13 @@ def Embedding(num_embeddings, embedding_dim, padding_idx):
     return m
 
 
-def LayerNorm(embedding_dim):
-    m = nn.LayerNorm(embedding_dim)
-    return m
+class LayerNorm(nn.LayerNorm):
+
+    def __init__(self, embedding_dim):
+        super().__init__(embedding_dim)
+
+    def forward(self, x):
+        return super().forward(x.float()).type_as(x)
 
 
 def Linear(in_features, out_features, bias=True):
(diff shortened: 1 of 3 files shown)


alexeib:
whats the speed difference?

myleott:
Good question. On 8 GPUs with `--update-freq 16` it's 188.7k wps vs 180.0k wps so almost 5% slower :/ So maybe it's not worth doing this...

alexeib:
what do you think about making this an option, like --safe-fp16? we can also check whether we need to do fp32 cast before the softmax and include that in that option if the difference is small?

myleott:
I'm going to abandon this:
- It's complicated to make this work automatically, since the inputs need to be converted to FP32 and the output converted back to FP16, which requires modifications in the Model anyway.
- We have FP32 to test whether there's a regression caused by --fp16, so we can always debug if needed.

I'm more inclined to add an --apex-fp16 option or something like that, since Nvidia has already put in quite a bit of work to make FP16 safe, no need to reinvent it. Last time I evaluated apex/amp it was slower than the fairseq version (specifically the FP16 optimizer) but I didn't dig into why.

PR #397, opened 2018-12-06
Add --fp16-scale-tolerance

myleott:
Summary: Let's only decrease the loss scale if a large enough percentage of batches overflow.

Differential Revision: D13355159


myleott:
Previously we reduced the loss scale if a single update overflowed. But in practice we may actually get faster/better convergence by using a bigger loss scale and accepting some percentage of overflows (i.e., we throw away 5% of batches, but the other 95% have a bigger loss scale).

PR #403, opened 2018-12-07
fix data checking report bug

pengli09:
The original code reports the size of a valid sample instead of an invalid one when raising an Exception , which will make people confused.

### DIFF
--- a/fairseq/data/data_utils.py
+++ b/fairseq/data/data_utils.py
@@ -100,12 +100,14 @@ def filter_by_size(indices, size_fn, max_positions, raise_exception=False):
 
     ignored = []
     itr = collect_filtered(check_size, indices, ignored)
+
+    if len(ignored) > 0 and raise_exception:
+        raise Exception((
+            'Size of sample #{} is invalid (={}) since max_positions={}, '
+            'skip this example with --skip-invalid-size-inputs-valid-test'
+        ).format(idx, size_fn(ignored[0]), max_positions))
+            
     for idx in itr:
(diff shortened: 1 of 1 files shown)


fairseq/data/data_utils.py
```diff
@@ -100,12 +100,14 @@ def check_size(idx):
 
     ignored = []
     itr = collect_filtered(check_size, indices, ignored)
+
+    if len(ignored) > 0 and raise_exception:
+        raise Exception((
+            'Size of sample #{} is invalid (={}) since max_positions={}, '
+            'skip this example with --skip-invalid-size-inputs-valid-test'
+        ).format(idx, size_fn(ignored[0]), max_positions))
```
myleott: This doesn't work, since collect_filtered is a generator, thus `ignored` will only be populated if the very first example is an invalid size.

fairseq/data/data_utils.py
```diff
@@ -100,12 +100,14 @@ def check_size(idx):
...
+            
     for idx in itr:
-        if len(ignored) > 0 and raise_exception:
-            raise Exception((
-                'Size of sample #{} is invalid (={}) since max_positions={}, '
-                'skip this example with --skip-invalid-size-inputs-valid-test'
-            ).format(idx, size_fn(idx), max_positions))
```
myleott: The correct change is to revert to the old code, but change this line to:
```
).format(ignored[0], size_fn(ignored[0]), max_positions))
```

PR #404, opened 2018-12-07
Improve memory efficiency of FP16 optimization

myleott:
Previously when training with --fp16, we stored a copy of the model parameters in FP32 for optimization, which consumed a lot of memory. An alternative is to just do the conversions to FP32 on the fly, which allows the caching allocator to reuse/save some memory.

This reduces peak memory usage by ~20% with a negligible reduction in training speed (~2% slower) when training a big transformer on 8 GPUs on wmt en-de with --update-freq=16.

This does not affect convergence, i.e., models will train exactly as they did before.

### DIFF
--- a/fairseq/optim/fp16_optimizer.py
+++ b/fairseq/optim/fp16_optimizer.py
@@ -10,7 +10,7 @@ import torch
 from fairseq import optim, utils
 
 
-class DynamicLossScaler:
+class DynamicLossScaler(object):
 
     def __init__(self, init_scale=2.**15, scale_factor=2., scale_window=2000):
         self.loss_scale = init_scale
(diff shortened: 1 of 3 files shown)


mcarilli:
I was led here by https://github.com/NVIDIA/apex/issues/139.  I'm confused about this, so please correct me if I have misinterpreted anything. 

Near as I can tell, what you're doing is:
```
fp16 param -> fp32 param
fp16 grad -> fp32 grad
fp32 param + fp32 grad -> fp32 param -> fp16 param
discard fp32 stuff
```
I believe this is an unsafe application of the mixed precision recipe. You carry out updates in fp32, but then cast back to fp16, which loses all the additional mantissa information of the master params. This additional mantissa information is (at least, according to the concept of the mixed precision recipe) essential to capture small updates that accumulate over time (in other words, updates that are not big enough to be reflected in fp16 for a single step, but eventually get there as they accumulate over multiple steps).

Moreover, since half-precision addition kernels in Pytorch internally upconvert to fp32, perform the addition, then downconvert the output back to fp16, the kernels are internally implementing the exact same sequence you've rigged, with a lot less memory traffic.

My interpretation is, if your model still trains to the desired accuracy with this PR, the takeaway is that you don't need master params at all for your use case, and might as well just train directly with the fp16 params.  I definitely think it's worth a try.  It may work and you may save yourself a decent amount of performance (at least 2%, probably more since you won't need to do any manual fp16<->fp32 copies and the memory traffic required for the gradient accumulates will be halved).

myleott:
Your interpretation seems right. The benefit of this over pure FP16 optimization is that the optimizer state (e.g., momentum buffers) are kept in FP32. In the past when I tried pure FP16 optimization I observed poor convergence.

mcarilli:
Interesting.  I guess it's less surprising that you would observe this behavior with Adam rather than with vanilla SGD.

Issue #438, opened 2019-01-09
Extend TranslationTask, Dictionary and architecture


davidecaroselli:
Hi fairseq team!

I'm working on extending fairseq with my custom architecture.
Basically I created a custom Dictionary subclass (supporting bpe-encoding), and I want to add my custom TranslationTask subclass that knows how to load such model.
I also want to be able to declare my custom architecture.

At the moment I don't see a way to do that without changing the code (preprocess.py, train.py, interactive.py ecc..). The main problems I see are two:

1. There is no way to include a custom module that register new architectures or tasks. Something similar to tensor2tensor's --t2t_usr_dir.
2. The pre-process script make the assumption that the only available Dictionary is the standard one, and it does not ask the task to provide its implementation.

So I would like to ask if there is some option I missed and this is actually already possible or, if not, would you accept a pull request enabling this feature?

At the moment I ended up cloning all python scripts in fairseq home folder.. and this is definitely a bad idea!

Thanks!

myleott:
Interesting! In the past we've actually recommended forking train.py and putting the imports there, but I like this `--user-dir` idea; a PR for this would be great! Otherwise, we're planning to do a kind of fairseq fix/hackathon thing at the end of the month and we can build it then.

Re: preprocess script and asking the task, I'm a bit uneasy about adding more complexity to the task API. Might it be simpler to just make this configurable in preprocess.py directly? Alternatively, if your changes to Dictionary are general enough could we just pull them into the existing base class?

davidecaroselli:
Cool, I will work on this `--user-dir` and send you a pull request asap! I would also like to edit the setup script you have, in order to let preprocess.py and train.py be globally available in the cli, just like t2t-train and t2t-datagen. This way once installed fairseq, we could simply run something like:
```bash
fairseq-preprocess -a ... -b ...
fairseq-train -a ... -b ...
```
Without the need to keep the fairseq github repo somewhere. Does it sounds reasonable to you?

On the preprocess, I understand that the "change" is not that immediate. Let me try to explain my point. The current implementation make the assumption that the input text (for both training and decoding) is already tokenized even with bpe. On top of that you have a strict definition of what a Dictionary is, on how to train it and how it stores its files. 

The problem with this is that is very rigid, and does not allow to choose a different tokenization/numerization. For example a bpe model has its own tokens and its own `token -> id` map, and it is not possible to use it with fairseq. A more elastic structure could be something more similar to what t2t does:

[...]

PR #448, opened 2019-01-14
Support custom Dictionary implementations in 'preprocess.py'

davidecaroselli:
The `preprocess.py` script has been refactored in order to:

1. Use the `options` module for command line arguments  parsing. This will give to `preprocess.py` the ability to load custom modules with `--user-dir` flag (already implemented to all other binaries)
2. Dictionary loading and building code has moved to Task implementation. This allows custom Dictionary classes to be used during the data generation step.

### DIFF
--- a/fairseq/tasks/translation.py
+++ b/fairseq/tasks/translation.py
@@ -94,18 +94,6 @@ class TranslationTask(FairseqTask):
         self.tgt_dict = tgt_dict
 
     @classmethod
-    def load_dictionary(cls, filename):
-        return Dictionary.load(filename)
-
-    @classmethod
-    def build_dictionary(cls, filenames, workers=1, threshold=-1, nwords=-1, padding_factor=8):
-        d = dictionary.Dictionary()
-        for filename in filenames:
-            Tokenizer.add_file_to_dictionary(filename, d, tokenizer.tokenize_line, workers)
-        d.finalize(threshold=threshold, nwords=nwords, padding_factor=padding_factor)
-        return d
-
-    @classmethod
     def setup_task(cls, args, **kwargs):
         """Setup the task (e.g., load dictionaries).
 
--- a/fairseq/tasks/translation.py
+++ b/fairseq/tasks/translation.py
@@ -8,7 +8,7 @@
 import itertools
 import os
 
-from fairseq import options, utils
+from fairseq import options, utils, tokenizer
 from fairseq.data import (
     ConcatDataset,
     data_utils,
@@ -17,7 +17,8 @@ from fairseq.data import (
     IndexedDataset,
     IndexedRawTextDataset,
     LanguagePairDataset,
-)
+    dictionary)
+from fairseq.tokenizer import Tokenizer
 
 from . import FairseqTask, register_task
 
@@ -93,6 +94,18 @@ class TranslationTask(FairseqTask):
         self.tgt_dict = tgt_dict
 
     @classmethod
+    def load_dictionary(cls, filename):
+        return Dictionary.load(filename)
+
+    @classmethod
+    def build_dictionary(cls, filenames, workers=1, threshold=-1, nwords=-1, padding_factor=8):
+        d = dictionary.Dictionary()
+        for filename in filenames:
+            Tokenizer.add_file_to_dictionary(filename, d, tokenizer.tokenize_line, workers)
+        d.finalize(threshold=threshold, nwords=nwords, padding_factor=padding_factor)
+        return d
+
+    @classmethod
     def setup_task(cls, args, **kwargs):
         """Setup the task (e.g., load dictionaries).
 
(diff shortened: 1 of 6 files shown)


davidecaroselli:
Hi @myleott , any plans to include this PR on master branch? It is fundamental to provide our own task/dictionary implementation. Do you need some changes to be more inline with your implementation idea? Just let me know!

myleott:
Hey, sorry for the delay! I'm just looking this over now and it looks pretty good.

To make sure I have the right context, you mentioned previously [1] that the custom dictionary is to support BPE, right? Currently we require the user to do this prior to calling preprocess.py, as is done in the various `prepare-*.sh` example scripts [2], and then in fairseq we assume whitespace-tokenized inputs/outputs. We also provide a few helpers for removing certain kinds of BPE automatically in generate.py, e.g., `--remove-bpe=sentencepiece`.

Is the proposal to first make Dictionary configurable throughout (basically this PR), and then have a second PR to move the existing whitespace-tokenization logic into the Dictionary (i.e., delete the Tokenizer class)?

[1] https://github.com/pytorch/fairseq/issues/438#issuecomment-453159180
[2] https://github.com/pytorch/fairseq/tree/master/examples/translation

fairseq/tasks/translation.py
```diff
@@ -92,6 +93,18 @@ def __init__(self, args, src_dict, tgt_dict):
...
+    def load_dictionary(cls, filename):
+        return Dictionary.load(filename)
+
+    @classmethod
+    def build_dictionary(cls, filenames, workers=1, threshold=-1, nwords=-1, padding_factor=8):
+        d = dictionary.Dictionary()
+        for filename in filenames:
+            Tokenizer.add_file_to_dictionary(filename, d, tokenizer.tokenize_line, workers)
+        d.finalize(threshold=threshold, nwords=nwords, padding_factor=padding_factor)
+        return d
```
myleott: Let's move these implementations into the base class for now, so that LanguageModelTask (and some other internal Task variants) will continue to work.

preprocess.py
```diff
@@ -9,71 +9,29 @@
...
 from itertools import zip_longest
 import os
 import shutil
 
-
-from fairseq.data import indexed_dataset, dictionary
-from fairseq.tokenizer import Tokenizer, tokenize_line
+from fairseq import options
+from fairseq.data import indexed_dataset
+from fairseq.tasks import TASK_REGISTRY
```
myleott: Let's add a simple `get_task` method to `fairseq/tasks/__init__.py` to make this import cleaner:
```
def get_task(name):
    return TASK_REGISTRY[name]
```

preprocess.py
```diff
@@ -90,49 +48,45 @@ def dict_path(lang):
...
-        assert not args.srcdict, "cannot combine --srcdict and --joined-dictionary"
-        assert not args.tgtdict, "cannot combine --tgtdict and --joined-dictionary"
-        src_dict = build_dictionary(
-            {train_path(lang) for lang in [args.source_lang, args.target_lang]},
-            args.workers,
-        )
+        if args.srcdict:
+            src_dict = task.load_dictionary(args.srcdict)
+        elif args.tgtdict:
+            src_dict = task.load_dictionary(args.tgtdict)
```
myleott: Cool, this is a nice feature, but let's also make sure the user doesn't specify both srcdict and tgtdict in this case:
```
assert not args.srcdict or not args.tgtdict, "cannot use both --srcdict and --tgtdict with --joined-dictionary"
```

preprocess.py
```diff
@@ -300,12 +235,6 @@ def consumer(tensor):
     return res
 
 
-def binarize_with_load(args, filename, dict_path, output_prefix, lang, offset, end):
```
myleott: Thanks for removing these, they somehow crept in from other internal users, but should be moved to those call sites directly since they aren't used in fairseq.

preprocess.py
```diff
@@ -90,49 +48,45 @@ def dict_path(lang):
         return dest_path("dict", lang) + ".txt"
```
myleott: Let's add a local helper function to build the dictionary, which can include all the extra options (currently missing below), such as threshold, nwords, padding_factor, etc.
```
def build_dictionary(filenames, src=False, tgt=False):
    assert src ^ tgt
    return task.build_dictionary(
        filenames,
        workers=args.workers,
        threshold=args.thresholdsrc if src else args.thresholdtgt,
        nwords=args.nwordssrc if src else args.nwordstgt,
        padding_factor=args.padding_factor,
    )
```

preprocess.py
```diff
@@ -90,49 +48,45 @@ def dict_path(lang):
...
-        )
+        if args.srcdict:
+            src_dict = task.load_dictionary(args.srcdict)
+        elif args.tgtdict:
+            src_dict = task.load_dictionary(args.tgtdict)
+        else:
+            assert (
+                args.trainpref
+            ), "--trainpref must be set if --srcdict is not specified"
+            src_dict = task.build_dictionary(
```
myleott: This becomes `src_dict = build_dictionary(..., src=True)` and similarly below.

davidecaroselli:
> Is the proposal to first make Dictionary configurable throughout (basically this PR), and then have a second PR to move the existing whitespace-tokenization logic into the Dictionary (i.e., delete the Tokenizer class)?

Yes, that's right! I will work on the requested changes, thanks for the feedback.

fairseq/tasks/translation.py
davidecaroselli: Done with commit https://github.com/pytorch/fairseq/pull/448/commits/03f0f3008a4dade4fd2330d9b34eb9023139a072

preprocess.py
davidecaroselli: Done https://github.com/pytorch/fairseq/pull/448/commits/b73bb739f94fb3db17a6ba46329765a9642143a8

preprocess.py
davidecaroselli: Done https://github.com/pytorch/fairseq/pull/448/commits/037aae29e44e2bba41ef3ee5d858e8575aa6c249

davidecaroselli:
Requested changes completed, green light from test cases too!

myleott:
Looks good! I'll merge this shortly, just need to wait for internal tests

PR #449, opened 2019-01-15
FIX: '--user-dir' on multi-gpu

davidecaroselli:
On a multi-gpu training scenario, the `train.py` script spawns new processes with `torch.multiprocessing.spawn`. Unfortunately those child processes don't inherit the modules imported with `--user-dir`.

This pull request fixes this problem: custom module import in now explicit on every `main()` function.

### DIFF
--- a/generate.py
+++ b/generate.py
@@ -25,8 +25,7 @@ def main(args):
     assert args.replace_unk is None or args.raw_text, \
         '--replace-unk requires a raw text dataset (--raw-text)'
 
-    if args.user_dir is not None:
-        import_user_module(args.user_dir)
+    import_user_module(args)
 
     if args.max_tokens is None and args.max_sentences is None:
         args.max_tokens = 12000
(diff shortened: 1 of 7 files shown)


generate.py
```diff
@@ -24,6 +25,9 @@ def main(args):
     assert args.replace_unk is None or args.raw_text, \
         '--replace-unk requires a raw text dataset (--raw-text)'
 
+    if args.user_dir is not None:
+        import_user_module(args.user_dir)
```
myleott: Please add this to eval_lm.py as well

davidecaroselli: Sure!

davidecaroselli: Done!

myleott:
Unit tests seem to be failing on this. Please make sure to run `python setup.py test`.

davidecaroselli:
Hi @myleott 

just pushed the fix. Now there are no errors related to this. I'm not able to run all test cases due some other random problems (device, cuda and so on) so please, double check.

I didn't expect unit testing was using also the "binaries" or argument parsing, but it should work!

Thanks

PR #460, opened 2019-01-18
Enforce UTF-8 when open() text files

davidecaroselli:
When opening text files without specifying the encoding (i.e. `open(path, "r")` or `open(path, "w")`), python3 will use the preferred locale encoding (`locale.getpreferredencoding()`) so the result is platform dependent and can change from one machine to another.

I believe fairseq should enforce its standard (UTF-8 seems like the best choice to me). This pull request explicity specify UTF-8 encoding when reading text files.


myleott:
I think this makes sense. I did some light testing and didn't notice any issues, so let's merge it and revisit if anything pops up later.

PR #467, opened 2019-01-24
 change f"{args}" to "{}".format(args)

vufg:
Although both are supported by Python 3.6, I think it would be better to unify the usage of string format function.


myleott:
Thanks! We should really do this everywhere, since it's the only reason we don't support Python 3.5. Let's start with this though.

Issue #477, opened 2019-01-28
Bug in shebang for Linux


davidecaroselli:
The shebang line in fairseq binaries are broken in Linux:
```bash
#!/usr/bin/env python3 -u
```

The `command [arg]` syntax is not valid on Linux, as already explained here for example: [StackOverflow - Shebang line with `#!/usr/bin/env command --argument` fails on Linux
](https://unix.stackexchange.com/questions/63979/shebang-line-with-usr-bin-env-command-argument-fails-on-linux).

I'm creating a PR together with feature discussed in #450 to solve this.

frankang:
I'm just curious that the in link you posted above mentioned that an error will be raised when executing that script. 
` /usr/bin/env: ruby --verbose: No such file or directory`
however I didn't see error output when run fairseq with `#!/usr/bin/env python3 -u `
How could this happen if shebang interpret the argument as "python3 -u"

myleott:
Yeah, I also have not observed any issues with this. @davidecaroselli, does it cause that error for you?

We added the `-u` to make Python's output unbuffered, since we log training progress to stdout. It's not critical to have it, but seemed to improve logging latency.

davidecaroselli:
In order to replicate the bug (tested on Ubuntu 16.04 and 18.04):

1. Create a simple "hello world" python script:
```bash
echo -e '#/usr/bin/env python3 -u python3 -u\nprint("hello world")' > test
```
2. Try the script:
```bash
$ python3 test
hello world
```
3. Now make the script executable:
```bash
chmod +x test
```
4. Try to run the script:
```bash
$ ./test 
/usr/bin/env: ‘python3 -u’: No such file or directory
```

myleott:
Fixed by https://github.com/pytorch/fairseq/commit/829bd8ce5fc336991c828f178ac5f231250ef77d

Issue #450, opened 2019-01-16
Group binaries in "bin" dir and install them globally with setup.py


davidecaroselli:
Hi fairseq team!

I would like to make a pull-request aiming to give fairseq a group of globally callable binaries. Just like tensro2tensor's `t2t-train` or `t2t-datagen`, I think that fairseq should expose its binaries like: `fairseq-datagen`, `fairseq-train`, `fairseq-lmeval`, `fairseq-generate` and `fairseq-decode`. This is also a fundamental step towards installable `pip` wheel.

So my idea is to:

1. Create a `bin` directory at top-level.
2. Move `eval_lm.py`, `preprocess.py`, `train.py`, `generate.py` and `interactive.py` into `bin/fairseq-lmeval`, `bin/fairseq-datagen`, `bin/fairseq-train`, `bin/fairseq-generate` and `bin/fairseq-decode`
3. Configure `setup.py` to make it install bins in user bin folder (I think usually `/usl/local/bin`)
4. For the sake of back-compatibility, replace top-level scripts (`eval_lm.py`, `preprocess.py`, ...) with scripts that print a deprecation warning and then run the correct `bin` script.

The result would be a much more integrated `fairseq` installation, for example I can run a training no matter from with the command:
```bash
fairseq-train  data-bin --save-dir ...
```

Are you interested in this feature?

myleott:
Fixed by https://github.com/pytorch/fairseq/commit/829bd8ce5fc336991c828f178ac5f231250ef77d

PR #478, opened 2019-01-28
Group binaries in "bin" dir and install them globally with setup.py

davidecaroselli:
Following discussion on #450, here all `fairseq-*` binaries are installed globally by `setup.py` and can be used without knowing the absolute path of the command.
This is also fundamental as a pre-condition to wheel distribution.

It also fix bug #477


myleott:
Thanks for doing this! This will take a bit of time to review, since we use some of these binaries internally and we'll need to update those call sites at the same time as we merge this.

If needed, is it possible for me to push small changes to this branch? I guess I can make a pull request to davidecaroselli/fairseq...

davidecaroselli:
@myleott sure! I think you can simply send pull request to my `features/global-binaries` branch, and once I accept them, they should popup here immediately.

> since we use some of these binaries internally

yeah, I expected this. I have maintained those binaries for back-compatibility, the idea is to avoid the need to update external tools that rely on them. But if you were using those bins as python modules (just like test cases did), you are forced to update those.

On the bright side, the change I made, moving the "logic" of the binaries in the main repo, goes in this direction: those functionalities are internal to the fairseq core, and binaries only refer them. I personally think this is a more solid implementation; hope you agree.

PR #490, opened 2019-02-06
Look for bytecode(.pyc) files when importing modules

akinh:
Summary:
We currently have 47 broken tests in pytorch_translate (https://fburl.com/tests/0riyt9ly) because when we run tests in mode/opt, machine only sees compiled files with *.pyc extension instead of original *.py files.

Therefore, all of these imports are failing with "key not found" errors (https://our.intern.facebook.com/intern/tests/details/562949963428218/debug/1561945).

Therefore, changing the way we are finding files, so that it includes compiled python files as well. This was the least intrusive way of solving this issue, let me know if you have other ideas.

Differential Revision: D13969981


myleott:
Spoke offline, this is not needed.

PR #495, opened 2019-02-09
Add fairseq to PyPI

myleott:
- fairseq can now be installed via pip: `pip install fairseq`
- command-line tools are globally accessible: `fairseq-preprocess`, `fairseq-train`, `fairseq-generate`, etc.


davidecaroselli:
That's great!
On [PyPI](https://pypi.org/project/fairseq/#history) I see that there is version 0.6.1, but this is not available on this github repo (I suggest to keep them "linked", so that it is easy for someone to get the source code of the pip install).

Which commit correspond to version 0.6.1?

myleott:
Good call, it's this commit that corresponds to 0.6.1. I'll tag the release now.

PR #513, opened 2019-02-19
moving masking logic to collate

rutyrinott:
Summary: Move masking logic to data_utils

Reviewed By: kartikayk, jingfeidu

Differential Revision: D14098403

### DIFF
--- a/fairseq/data/data_utils.py
+++ b/fairseq/data/data_utils.py
@@ -7,8 +7,48 @@
 
 import contextlib
 import os
-
+import math
 import numpy as np
+
+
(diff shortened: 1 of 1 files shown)


alexeib:
why not do this during data preprocessing stage, i think thats what bert does, right? i've done some preliminary bert data processing reimplementation here: https://github.com/fairinternal/fairseq-py/blob/bi_trans_lm/scripts/data/prepare_sentence_samples.py although it doesnt have the masking part yet

myleott:
Going to merge this in #516 to keep the sync active.

PR #548, opened 2019-03-01
Fixed the issue that no space in string converted from tensor

lfthwjx:


myleott:
Thanks, can you also change this to make `--remove-bpe=sentencepiece` work:
https://github.com/pytorch/fairseq/blob/f296824f4013dc28b471c47d7779547460afc7f0/fairseq/data/data_utils.py#L220
to:
```
sentence = sentence.replace(' ', '').replace('\u2581', ' ').strip()
```

lfthwjx:
Changed, thanks

Issue #517, opened 2019-02-21
Importing fairseq causes joblib to crash


louismartin:
I have a weird bug that causes joblib parallel jobs to crash when any of my script imports fairseq.

I have made a minimal repro:

Script `test.py`:
```
import fairseq


def test():
    pass
```

Script `main.py`:
```
from joblib import Parallel, delayed

from test import test


if __name__ == '__main__':
    Parallel(n_jobs=2)([delayed(test)()])
```

Error when running `python main.py`:
```
joblib.externals.loky.process_executor._RemoteTraceback:
'''
Traceback (most recent call last):
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/site-packages/joblib/externals/loky/process_executor.py", line 391, in _process_worker
    call_item = call_queue.get(block=True, timeout=timeout)
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/multiprocessing/queues.py", line 113, in get
    return _ForkingPickler.loads(res)
  File "/private/home/louismartin/tmp/fairseq_repro/test.py", line 1, in <module>
    import fairseq
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/site-packages/fairseq/__init__.py", line 8, in <module>
    from .multiprocessing_pdb import pdb
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/site-packages/fairseq/multiprocessing_pdb.py", line 14, in <module>
    class MultiprocessingPdb(pdb.Pdb):
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/site-packages/fairseq/multiprocessing_pdb.py", line 20, in MultiprocessingPdb
    _stdin_fd = sys.stdin.fileno()
AttributeError: 'NoneType' object has no attribute 'fileno'
'''

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "main.py", line 5, in <module>
    Parallel(n_jobs=2)([delayed(test)()])
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/site-packages/joblib/parallel.py", line 934, in __call__
    self.retrieve()
  File "/private/home/louismartin/miniconda3/envs/test/lib/python3.7/site-packages/joblib/parallel.py", line 833, in retrieve
    self._output.extend(job.get(timeout=self.timeout))
[...]

PR #553, opened 2019-03-04
Try to access sys.stdin.fileno() only at runtime and not during import

louismartin:
Accessing sys.stdin.fileno() raises an error in multiple contexts
(pytest, joblib, jupyter...).
Thus accessing it at the top level of the file can cause other scripts
to crash when they import fairseq.
This is why it is moved inside the method of MultiprocessingPdb to only
be accessed at runtime if needed.

See  Issue #517

### DIFF
--- a/fairseq/pdb.py
+++ b/fairseq/pdb.py
@@ -16,6 +16,10 @@ __all__ = ['set_trace']
 
 _stdin = [None]
 _stdin_lock = multiprocessing.Lock()
+try:
+    _stdin_fd = sys.stdin.fileno()
+except Exception:
+    _stdin_fd = None
 
 
 class MultiprocessingPdb(pdb.Pdb):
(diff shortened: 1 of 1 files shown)


fairseq/pdb.py
```diff
@@ -14,7 +14,6 @@
 __all__ = ['set_trace']
 
 
-_stdin_fd = sys.stdin.fileno()
```
myleott:
```
try:
    _stdin_fd = sys.stdin.fileno()
except:
    _stdin_fd = None
```

fairseq/pdb.py
```diff
@@ -33,7 +32,7 @@ def _cmdloop(self):
         with _stdin_lock:
             try:
                 if not _stdin[0]:
-                    _stdin[0] = os.fdopen(_stdin_fd)
+                    _stdin[0] = os.fdopen(stdin_bak.fileno())
                 sys.stdin = _stdin[0]
                 self.cmdloop()
```
myleott:
```
if _stdin_fd is not None:
    if not _stdin[0]:
        _stdin[0] = os.fdopen(_stdin_fd)
    sys.stdin = _stdin[0]
self.cmdloop()
```

louismartin:
Took your comments into account, works great on my side.

Issue #574, opened 2019-03-13
Huge amount of CPU RAM needed during training


davidecaroselli:
Hello team!

With the current version of fairseq we noticed that a huge amount of RAM (CPU RAM, not GPU RAM) is required in order to run the training. Moreover this is correlated to the number of GPU used on the same machine.

So my guess is that the binarized data used for the training is completely loaded in RAM for every GPU process, this will result in having the amount of CPU RAM is roughly:
`RAM ~= (number of GPUs) * sizeof(binarized data)`.
If this is true, the amount of RAM needed for medium/large training sets is huge (hundreds of GB) wrt size of training set (less than 100 GB).

If this is the case, why can't we use a memory mapped training set? So that the amount of RAM depends exclusively on `sizeof(binarized data)` ?

I'm available to work on this if needed, can you please give me some code context, or a good starting point to begin with?

myleott:
One option is to use `--fix-batches-to-gpus`, which will load only 1/N of the data in each process, assuming N workers.

mmapping the whole training set per node is reasonable as well and may have some interesting tradeoffs compare to the `--fix-batches-to-gpus` option.

davidecaroselli:
From the documentation of `--fix-batches-to-gpus` I understand that there could be some problems with translation quality (am I correct?), but I don't really understand why.. If the whole cluster should work together to train on the training set **without** repetitions, why splitting batches per-GPU will result in problematic training?

In other words: which is the advantage to have duplicated batches per GPU? If a batch "k" is run on "GPU X", for sure won't be seen by "GPU Y" untile next epoch, is that correct?
If this is the case, why preventing batch "k" to be replicated in all GPUs != "GPU X" should be a problem?

Thanks for the info @myleott ! I just want to understand better how fairseq work in order to be more confident into changing the training DataSet implementation.

myleott:
When using `--fix-batches-to-gpus` then it will slightly reduce randomness, which can (very slightly) affect model quality if training on many GPUs.

This is because we will never shuffle batches across GPUs, only within a GPU.

Consider the normal case, training with 2 GPUs on a dataset with 4 batches. You might have something like:
```
             GPU 1     GPU 2
Epoch 0.0   batch 1   batch 2
Epoch 0.5   batch 3   batch 4
Epoch 1.0   batch 1   batch 3
Epoch 1.5   batch 4   batch 2
```

[...]

PR #589, opened 2019-03-20
Memory-Mapped IndexedDataset implementation

davidecaroselli:
Following discussion in https://github.com/pytorch/fairseq/issues/574:

 - Implemented MMapIndexedDataset and MMapIndexedDatasetBuilder compatible with IndexedDataset/IndexedDatasetBuilder
- Update scripts/read_binarized.py to support new MMapIndexedDataset
- Option '--raw-text' and '--lazy-load' replaced with '--dataset-impl' and moved the option definition custom task args to more high-level options.add_dataset_args() (more appropriate)
- Implemented also utils functions in indexed_dataset: make_dataset(), dataset_exists()


fairseq/tasks/language_modeling.py
```diff
@@ -69,10 +66,6 @@ def add_args(parser):
                                  'If set to "eos", includes only one sentence per sample.')
         parser.add_argument('--tokens-per-sample', default=1024, type=int,
                             help='max number of tokens per sample for LM dataset')
-        parser.add_argument('--lazy-load', action='store_true',
-                            help='load the dataset lazily')
-        parser.add_argument('--raw-text', default=False, action='store_true',
-                            help='load raw text dataset')
```
myleott: Let's provide backward compatibility for these. Add back these flags, change the description to indicate `[deprecated]` and add logic here: https://github.com/pytorch/fairseq/blob/master/fairseq/options.py#L111-L115.

Something like:
```
if getattr(args, 'raw_text', False):
    utils.deprecation_warning('--raw-text is deprecated, please use --dataset-impl=raw')
    args.dataset_impl = 'raw'
elif getattr(args, 'lazy_load', False):
    utils.deprecation_warning('--lazy-load is deprecated, please use --dataset-impl=lazy')
    args.dataset_impl = 'lazy'
```

davidecaroselli: I still have to include these changes, I will work on this asap!

fairseq/data/indexed_dataset.py
```diff
@@ -262,3 +293,169 @@ def finalize(self, index_file):
         write_longs(index, self.data_offsets)
         write_longs(index, self.sizes)
         index.close()
+
+
+def _warmup_mmap_file(path):
```
ebetica: Hey @davidecaroselli, I noticed in your implementation, you require a full read of the dataset at the beginning of training to warm up the mmap file. Is this actually useful? Can I get away with not warming up the mmap file?

davidecaroselli: Hi! Yes, that is super useful because that warmup ensure best performance during actual training. However it is not mandatory, meaning that everything works even without it. So if, for some strange reason, you want to avoid it, just comment the call!

PR #598, opened 2019-03-26
Update data_utils.py for 

yashkumaratri:
Correcting the syntax error in assert function cause of a character before error message.

Assertion and the code is working fine now, Tested with wmt-ende task.

### DIFF
--- a/fairseq/data/data_utils.py
+++ b/fairseq/data/data_utils.py
@@ -156,7 +156,7 @@ def batch_by_size(
     for idx in indices:
         sample_lens.append(num_tokens_fn(idx))
         sample_len = max(sample_len, sample_lens[-1])
-        assert sample_len <= max_tokens, f"sentence at index {idx} exceeds max_tokens limit!"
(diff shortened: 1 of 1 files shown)


fairseq/data/data_utils.py
```diff
@@ -156,7 +156,7 @@ def is_batch_full(num_tokens):
     for idx in indices:
         sample_lens.append(num_tokens_fn(idx))
         sample_len = max(sample_len, sample_lens[-1])
-        assert sample_len <= max_tokens, f"sentence at index {idx} exceeds max_tokens limit!"
+        assert sample_len <= max_tokens, "sentence at index {idx} exceeds max_tokens limit!"
```
myleott: You're missing the format() call. Please change this to:
```
assert sample_len <= max_tokens, \
    "sentence at index {idx} exceeds max_tokens limit!".format(idx=idx)
```

yashkumaratri: Done

PR #647, opened 2019-04-21
reduce memory footprint for average_checkpoints

yqwangustc:
Summary:
the current implementation of average_checkpoints requires loading all
the model parameters into memory and then do the averaging. To average large
models (e.g., transformer) over a large number of checkpoints (e.g., >50),
it may require over 100GB memory.

Loading all the parameters is not necessary, as we know the number of models in advance.

Differential Revision: D15027513

### DIFF
--- a/scripts/average_checkpoints.py
+++ b/scripts/average_checkpoints.py
@@ -21,6 +21,8 @@ def average_checkpoints(inputs):
     params_dict = collections.OrderedDict()
     params_keys = None
     new_state = None
+    num_models = len(inputs)
+
     for f in inputs:
         state = torch.load(
             f,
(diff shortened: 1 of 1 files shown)


scripts/average_checkpoints.py
```diff
@@ -44,20 +46,18 @@ def average_checkpoints(inputs):
...
-            if k not in params_dict:
-                params_dict[k] = []
             p = model_params[k]
             if isinstance(p, torch.HalfTensor):
                 p = p.float()
-            params_dict[k].append(p)
+            if k not in params_dict:
+                params_dict[k] = p
+            else:
+                params_dict[k] += p
```
edunov: We need to be careful here with shared parameters, I believe right now it is broken if you have shared encoder/decoder input embeddings (encoder.embed_tokens.weight and decoder.embed_tokens.weight) as they get updated several times

yqwangustc: Thanks for pointing out this. Could you kindly give me an example of shared parameters so that I can better unittest this code ?

yqwangustc: See pr #747

PR #627, opened 2019-04-09
Fix generation with --no-early-stop

mryab:
Because the size of `unfinalized_scores` is equal to current `bsz` and not initial batch size, we need to index it by `unfin_idx` instead of `sent` in `is_finished`.
Fixes #588.

PR #651, opened 2019-04-22
Update sequence_generator.py

zhawe01:
Fix index out of bounds when enable --no-early-stop.


myleott:
Dupe of #627, merging that one instead.

PR #696, opened 2019-05-02
allowing sharded dataset

ngoyal2707:
Co-authored-by: myleott <myleott@fb.com>

Changing `data` to be `str` with colon separated list for loading sharded datasets. This change is useful for loading large datasets that cannot fit into, memory. The large dataset can be sharded and then each shard is loaded in one epoch in roudrobin manner.

For example, if there are `5` shards of data and `10` epochs then the shards will be iterated upon `[0, 1, 2, 3, 4, 0, 1, 2, 3, 4]`.

@myleott We need to look into `translation.py` as it currently already expects a list and then concats the datasets.

### DIFF
--- a/fairseq/tasks/cross_lingual_lm.py
+++ b/fairseq/tasks/cross_lingual_lm.py
@@ -42,7 +42,8 @@ class CrossLingualLMTask(FairseqTask):
     @staticmethod
     def add_args(parser):
         """Add task-specific arguments to the parser."""
-        parser.add_argument('data', help='path to data directory')
+        parser.add_argument('data', help='colon separated path to data directories list, \
+                            will be iterated upon during epochs in round-robin manner')
         parser.add_argument('--tokens-per-sample', default=512, type=int,
                             help='max number of total tokens over all segments'
                                  ' per sample')
@@ -106,12 +107,16 @@ class CrossLingualLMTask(FairseqTask):
 
         return cls(args, dictionary)
 
-    def _load_single_lang_dataset(self, split):
+    def _load_single_lang_dataset(self, split, epoch):
         loaded_datasets = []
+
+        paths = self.args.data.split(':')
+        assert len(paths) > 0
+        data_path = paths[epoch % len(paths)]
 
         for k in itertools.count():
             split_k = split + (str(k) if k > 0 else '')
-            path = os.path.join(self.args.data, split_k)
+            path = os.path.join(data_path, split_k)
 
             if self.args.raw_text and IndexedRawTextDataset.exists(path):
                 ds = IndexedRawTextDataset(path, self.dictionary)
@@ -124,7 +129,7 @@ class CrossLingualLMTask(FairseqTask):
                 if k > 0:
                     break
                 else:
-                    raise FileNotFoundError('Dataset not found: {} ({})'.format(split, self.args.data))
+                    raise FileNotFoundError('Dataset not found: {} ({})'.format(split, data_path))
 
             # Since we append each block with the classification_token,
             # we need to effectively create blocks of length
@@ -136,7 +141,7 @@ class CrossLingualLMTask(FairseqTask):
                 )
             )
 
-            print('| {} {} {} examples'.format(self.args.data, split_k, len(loaded_datasets[-1])))
+            print('| {} {} {} examples'.format(data_path, split_k, len(loaded_datasets[-1])))
 
         if len(loaded_datasets) == 1:
             dataset = loaded_datasets[0]
@@ -147,7 +152,7 @@ class CrossLingualLMTask(FairseqTask):
 
         return dataset, sizes
 
-    def load_dataset(self, split, combine=False, **kwargs):
+    def load_dataset(self, split, epoch=0, combine=False, **kwargs):
         """Load a given dataset split.
         Args:
             split (str): name of the split (e.g., train, valid, test)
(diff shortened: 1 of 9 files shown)


fairseq/tasks/cross_lingual_lm.py
```diff
@@ -106,12 +107,17 @@ def setup_task(cls, args, **kwargs):
...
         return cls(args, dictionary)
 
-    def _load_single_lang_dataset(self, split):
+    def _load_single_lang_dataset(self, split, epoch):
         loaded_datasets = []
 
+        paths = self.args.data.split(':')
+        assert len(paths) > 0
+        data_path = paths[epoch % len(paths)]
+        print("| data_path", data_path)
```
myleott: let's remove this print, it's redundant with the print on line 145

fairseq/tasks/language_modeling.py
```diff
@@ -145,9 +147,14 @@ def load_dataset(self, split, combine=False, **kwargs):
 
         loaded_datasets = []
 
+        paths = self.args.data.split(':')
+        assert len(paths) > 0
+        data_path = paths[epoch % len(paths)]
+        print("| data_path", data_path)
```
myleott: same, remove print?

train.py
```diff
@@ -107,10 +110,38 @@ def main(args):
...
+        max_positions=max_positions,
+        ignore_invalid_inputs=True,
+        required_batch_size_multiple=args.required_batch_size_multiple,
+        seed=args.seed,
+        num_shards=args.distributed_world_size,
+        shard_id=args.distributed_rank,
+        num_workers=args.num_workers,
+    )
+    # don't fast-forward, if the data is sharded
+    if len(args.data.split(":")) > 0:
```
myleott: is this supposed to be >1? otherwise when does it == 0?

Separately, we should exit this function early when len(...) == 1.

train.py
```diff
@@ -153,7 +184,7 @@ def train(args, trainer, task, epoch_itr):
 
         num_updates = trainer.get_num_updates()
         if args.save_interval_updates > 0 and num_updates % args.save_interval_updates == 0 and num_updates > 0:
-            valid_losses = validate(args, trainer, task, epoch_itr, valid_subsets)
+            valid_losses = validate(args, trainer, task, epoch_itr, [first_valid])
```
myleott: revert these changes to valid subsets

train.py
```diff
@@ -107,10 +110,37 @@ def main(args):
...
+        max_positions=max_positions,
+        ignore_invalid_inputs=True,
+        required_batch_size_multiple=args.required_batch_size_multiple,
+        seed=args.seed,
+        num_shards=args.distributed_world_size,
+        shard_id=args.distributed_rank,
+        num_workers=args.num_workers,
+    )
+    # don't fast-forward, if the data is sharded
+    epoch_state_dict['iterations_in_epoch'] = 0
```
myleott: overriding the state_dict like this seems messy. Can you instead pass an epoch kwarg to get_batch_iterator, and pass it through to EpochBatchIterator (instead of the current behavior of always setting it to 0): https://github.com/pytorch/fairseq/blob/master/fairseq/data/iterators.py#L97

train.py
```diff
@@ -367,17 +397,20 @@ def load_checkpoint(args, trainer, epoch_itr):
...
-            except FileNotFoundError as e:
-                if k > 0:
-                    break
-                raise e
+def load_dataset_splits(args, task, splits, epoch):
+    for split in splits:
+        if split == args.train_subset:
+            task.load_dataset(args.train_subset, combine=True, epoch=epoch)
+        else:
+            for split in args.valid_subset.split(','):
```
myleott: This will clobber the split variable from above. I think it will still work, but makes it harder to follow the code. If it's easier can we split this into load_train_dataset and load_valid_datasets?

train.py
```diff
@@ -107,10 +107,35 @@ def main(args):
...
+        max_sentences=args.max_sentences,
+        max_positions=max_positions,
+        ignore_invalid_inputs=True,
+        required_batch_size_multiple=args.required_batch_size_multiple,
+        seed=args.seed,
+        num_shards=args.distributed_world_size,
+        shard_id=args.distributed_rank,
+        num_workers=args.num_workers,
+    )
+    epoch_itr.load_state_dict(epoch_state_dict)
```
myleott: I'm still not sure about hacking the state dict like this. All we need is to set the initial epoch value right? Maybe we can pass an epoch= kwarg to get_batch_iterator so that EpochBatchIterator won't default epoch to 0: https://github.com/pytorch/fairseq/blob/master/fairseq/data/iterators.py#L97

train.py
```diff
@@ -367,17 +399,20 @@ def load_checkpoint(args, trainer, epoch_itr):
...
-                if k > 0:
-                    break
-                raise e
+def load_dataset_splits(args, task, dataset_splits, epoch):
+    for dataset_split in dataset_splits:
+        if dataset_split == args.train_subset:
+            task.load_dataset(args.train_subset, combine=True, epoch=epoch)
+        else:
+            for valid_sub_split in args.valid_subset.split(','):
+                for k in itertools.count():
```
myleott: Let's remove the itertools.count loop, it seems to be redundant with the existing loop in the tasks: https://github.com/fairinternal/fairseq-py/blob/master/fairseq/tasks/translation.py#L137

train.py
```diff
@@ -366,17 +398,18 @@ def load_checkpoint(args, trainer, epoch_itr):
...
-                raise e
+def load_dataset_splits(args, task, dataset_splits, epoch):
+    for dataset_split in dataset_splits:
+        if dataset_split == args.train_subset:
+            task.load_dataset(args.train_subset, combine=True, epoch=epoch)
+        else:
+            for valid_sub_split in args.valid_subset.split(','):
+                try:
+                    task.load_dataset(valid_sub_split, combine=True, epoch=epoch)
+                except FileNotFoundError as e:
```
myleott: You can remove the try/except

PR #697, opened 2019-05-02
added masked_lm task

ngoyal2707:
Co-authored-by: jingfeidu <jingfeidu@fb.com>

1) Adding `masked_lm` task for BERT like training. Code mostly taken from @jingfeidu 's implementation.

2) Added `has_eos` option to `block_pair_dataset` for working with dataset that has been preprocessed with having `eos`.

Depends on: https://github.com/pytorch/fairseq/pull/696

### DIFF
--- a/fairseq/data/block_pair_dataset.py
+++ b/fairseq/data/block_pair_dataset.py
@@ -32,6 +32,8 @@ class BlockPairDataset(FairseqDataset):
             doc: respect document boundaries and each part of the pair should belong to on document
             none: don't respect any boundary and cut tokens evenly
         short_seq_prob: probability for generating shorter block pairs
+        doc_break_size: Size for empty line separating documents. Typically 1 if
+                        the sentences have eos, 0 otherwise.
     """
 
     def __init__(
@@ -42,6 +44,7 @@ class BlockPairDataset(FairseqDataset):
         block_size,
         break_mode="doc",
         short_seq_prob=0.1,
+        doc_break_size=1,
     ):
         super().__init__()
         self.dataset = dataset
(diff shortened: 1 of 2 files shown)


fairseq/data/block_pair_dataset.py
```diff
@@ -32,6 +32,8 @@ class BlockPairDataset(FairseqDataset):
             doc: respect document boundaries and each part of the pair should belong to on document
             none: don't respect any boundary and cut tokens evenly
         short_seq_prob: probability for generating shorter block pairs
+        has_eos:    flag for indicating if the sentence has been preprocessed
+                    to end with eos or not.
```
myleott: nit: formatting should be:
```
has_eos: flag for indicating ...
    to end with eos or not.
```

fairseq/data/block_pair_dataset.py
```diff
@@ -42,6 +44,7 @@ def __init__(
         block_size,
         break_mode="doc",
         short_seq_prob=0.1,
+        has_eos=True,
```
myleott: actually it would be cleaner to just take doc_break_size directly

fairseq/tasks/masked_lm.py
```diff
@@ -0,0 +1,143 @@
...
+                    raise FileNotFoundError('Dataset not found: {} ({})'.format(split, data_path))
+            with data_utils.numpy_seed(self.seed + k):
+                loaded_datasets.append(
+                    BlockPairDataset(
+                        ds,
+                        self.dictionary,
+                        ds.sizes,
+                        self.args.tokens_per_sample,
+                        break_mode=self.args.break_mode,
+                        has_eos=True,
```
myleott: and then move the break size logic here:
```
doc_break_size=1,  # we expect sentences to end in eos
```

PR #706, opened 2019-05-06
Fix semisupervised_translation task

maxdel:
Pass required "sample_key" argument to forward-backward call in semi-supervised task.


pipibjc:
Good catch. Thanks for the fix! Merged into master.

PR #711, opened 2019-05-06
bugfix: passing args

taineleau:


fairseq/data/masked_lm_dataset.py
```diff
@@ -228,7 +228,8 @@ def _collate(
 
                 # mask according to specified probabilities.
                 masked_blk_one, masked_tgt_one = self._mask_block(
-                    s["block_one"], self.mask_idx, self.pad_idx, token_range)
+                    s["block_one"], self.mask_idx, self.pad_idx, token_range,
+                    masking_ratio=self.masking_ratio, masking_prob=self.masking_prob)
```
myleott: please also add random_token_prob

fairseq/data/masked_lm_dataset.py
```diff
@@ -228,7 +228,10 @@ def _collate(
 
                 # mask according to specified probabilities.
                 masked_blk_one, masked_tgt_one = self._mask_block(
```
ngoyal2707: `_mask_block` also gets called below for the block 2. 
maybe we can just use the member variable directly in `_mask_block()` function and not accept them as args ?

PR #763, opened 2019-05-28
Fix Tensorboard Init

Michaelvll:
Fix the mismatching between the parameter fed into `SummaryWriter` and the API of the latest [tensorboardX](https://github.com/lanpa/tensorboardX/blob/3e35c9b5f85e8ceb0294532d9eb772341a04c097/tensorboardX/writer.py#L192), i.e. "log_dir" -> "logdir".

### DIFF
--- a/fairseq/progress_bar.py
+++ b/fairseq/progress_bar.py
@@ -233,7 +233,7 @@ class tensorboard_log_wrapper(progress_bar):
             return None
         if key not in self._writers:
             self._writers[key] = self.SummaryWriter(
-                logdir=os.path.join(self.tensorboard_logdir, key),
+                os.path.join(self.tensorboard_logdir, key),
             )
             self._writers[key].add_text('args', str(vars(self.args)))
             self._writers[key].add_text('sys.argv', " ".join(sys.argv))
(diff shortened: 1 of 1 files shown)


fairseq/progress_bar.py
```diff
@@ -225,15 +225,15 @@ def __init__(self, wrapped_bar, tensorboard_logdir, args):
...
+                  "please see README for using tensorboard. (e.g. pip install tensorboardX)")
             self.SummaryWriter = None
 
     def _writer(self, key):
         if self.SummaryWriter is None:
             return None
         if key not in self._writers:
             self._writers[key] = self.SummaryWriter(
-                log_dir=os.path.join(self.tensorboard_logdir, key),
+                logdir=os.path.join(self.tensorboard_logdir, key),
```
myleott: Won't this break things for anyone using an old version of tensorboardX? It's not elegant, but let's remove the kwarg so that it works either way.

Michaelvll: Thank you! You are absolutely right. I just removed the kwarg in the latest commit.

PR #766, opened 2019-05-30
Clarify mixed precision training support

khoa-ho:
Change the wording to avoid confusion. Mixed precision ensures both higher arithmetic throughput and numerical stability, not exactly synonymous to pure half-precision/FP16 training. Also add mentioning of tensor cores since older generation GPUs without tensor cores don't support true mixed precision training.


myleott:
Thanks :)

PR #800, opened 2019-06-12
swap the training order in semisupervised_translation

hxbai:
it is better to train denoising autoencoder first.
for unsupervised mt, it may have better backtranslation.
ref: https://github.com/facebookresearch/XLM/blob/master/train.py#L288-L298

### DIFF
--- a/fairseq/tasks/semisupervised_translation.py
+++ b/fairseq/tasks/semisupervised_translation.py
@@ -349,6 +349,12 @@ class SemisupervisedTranslationTask(MultilingualTranslationTask):
             agg_sample_size += sample_size
             agg_logging_output[logging_output_key] = logging_output
 
+        if self.lambda_denoising > 0.0:
+            for lang_pair in self.args.lang_pairs:
+                _, tgt = lang_pair.split('-')
+                sample_key = _get_denoising_dataset_key(lang_pair)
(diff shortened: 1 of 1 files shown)


hxbai:
Is it needed? @pipibjc

pipibjc:
This won't change how we update the model. Unlike XLM, which update the model after each `trainer.xx_step`, fairseq implementation only update the model after we accumulate the gradients from all three tasks after each batch, i.e. [here](https://github.com/pytorch/fairseq/blob/master/fairseq/trainer.py#L341) which is after the `[train_step](https://github.com/pytorch/fairseq/blob/master/fairseq/trainer.py#L258-L261)`.

pipibjc:
We also compared between 1) update the model after each task (XLM implementation), and 2) update the model after all three tasks (fairseq implementation), but we didn't find any significant difference. We can get comparable unsupervised fr-en unsupervised MT result in terms of BLEU.
The latter one is easier to implement under current fairseq framework, so we went with 2).

hxbai:
Thank you! sorry for my carelessness. My experiments also show comparable results.

pipibjc:
No worries. Thanks for bringing this up!

Hbenmazi:
Hi! Could you tell me how to use fairseq for unsupervised translation?

PR #815, opened 2019-06-19
Enhanced MMapIndexedDataset: less memory, higher speed

davidecaroselli:
I have made an upgrade to my previous implementation of MMapIndexedDataset, now:
- It uses up to **4 times less memory and disk space**
- Words per second is slightly improved thanks to less memory access


davidecaroselli:
Please ignore this, I have just created a much more clear pull request here: https://github.com/pytorch/fairseq/pull/816

PR #816, opened 2019-06-19
Enhanced MMapIndexedDataset: less memory, higher speed

davidecaroselli:
I have made an upgrade to my previous implementation of MMapIndexedDataset, now:
- It uses up to **4 times less memory and disk space**
- Words per second is slightly improved thanks to less memory access

### DIFF
--- a/fairseq/data/indexed_dataset.py
+++ b/fairseq/data/indexed_dataset.py
@@ -15,9 +15,16 @@ import torch
 from . import FairseqDataset
 
 
-def make_builder(out_file, impl):
+def __best_fitting_dtype(vocab_size=None):
+    if vocab_size is not None and vocab_size < 65500:
+        return np.uint16
+    else:
+        return np.int32
+
+
+def make_builder(out_file, impl, vocab_size=None):
     if impl == 'mmap':
-        return MMapIndexedDatasetBuilder(out_file)
+        return MMapIndexedDatasetBuilder(out_file, dtype=__best_fitting_dtype(vocab_size))
     else:
         return IndexedDatasetBuilder(out_file)
 
@@ -63,6 +70,7 @@ dtypes = {
     5: np.int64,
     6: np.float,
     7: np.double,
+    8: np.uint16
 }
 
 
@@ -143,7 +151,7 @@ class IndexedDataset(FairseqDataset):
     @staticmethod
     def exists(path):
         return (
-            os.path.exists(index_file_path(path)) and os.path.exists(data_file_path(path))
+                os.path.exists(index_file_path(path)) and os.path.exists(data_file_path(path))
         )
 
     @property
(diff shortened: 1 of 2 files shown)


myleott:
Nice!

fairseq/data/indexed_dataset.py
```diff
@@ -15,9 +15,16 @@
 from . import FairseqDataset
 
 
-def make_builder(out_file, impl):
+def __best_fitting_dtype(vocab_size=None):
+    if vocab_size is not None and vocab_size < 65500:
+        return np.uint16
+    else:
+        return np.int32
```
myleott: This makes sense. Is this the main cause of the memory/speed improvement?

fairseq/data/indexed_dataset.py
```diff
@@ -440,11 +448,11 @@ def __len__(self):
...
     def __getitem__(self, i):
         ptr, size = self._index[i]
-        tensor = torch.from_numpy(np.frombuffer(self._bin_buffer, dtype=self._index.dtype, count=size, offset=ptr))
-        if tensor.dtype == torch.int64:
-            return tensor
-        else:
-            return tensor.long()
+        np_array = np.frombuffer(self._bin_buffer, dtype=self._index.dtype, count=size, offset=ptr)
+        if self._index.dtype != np.int64:
+            np_array = np_array.astype(np.int64)
```
myleott: Does this change improve memory usage and/or speed as well?

davidecaroselli: This is due the fact that invoking `tensor.long()` on a `uint16` tensor was throwing an exception (tensor supports casting only on `int64`, `int32`, `uint8`, ... except `uint16`).
This way I bypass the problem!

fairseq/data/indexed_dataset.py
davidecaroselli: Yes, this is actual boost. I now have a model that is 1/4th of the original and this means a huge saving in memory, disk and also load time!

davidecaroselli:
I'm running the very last test where I train a net on 8GPU for 6 hours and test that the translation results are equivalent to the same net trained with old version of MMapIndexedDataset.

PS: I just noticed the new version is out 0.7.0. That's great, but I was really hoping to have this improvement in the release! Too late I guess... Should I have to wait ~3 months, or are you planning to have a more regular and fast release schedule? It would be great to see all new improvements available in less time, even more important in this fast-growing stage of the framework

myleott:
> Should I have to wait ~3 months, or are you planning to have a more regular and fast release schedule?

We'll try to release more regularly. I cut 0.7.0 because we have a major logging refactoring landing soon and I wanted to cut a release right before it. We'll probably move to 0.8.0 very soon once that's merged.

I can put this in a 0.7.1 as soon as it's merged.

davidecaroselli:
Thanks @myleott for the amazing product! :)

myleott:
0.7.1 is on pypi with this change!

davidecaroselli:
Thanks a million!

PR #875, opened 2019-07-12
Initializing mask as a tensor of ints (not long)

taylanbil:
Since mask really is a tensor of ints, this change should be mathematically
equivalent to the base.

On the other hand, this has performance implications for xla, hence the
pull request.

### DIFF
--- a/fairseq/utils.py
+++ b/fairseq/utils.py
@@ -172,8 +172,14 @@ def make_positions(tensor, padding_idx, onnx_trace=False):
 
     Position numbers begin at padding_idx+1. Padding symbols are ignored.
     """
+    # The series of casts and type-conversions here are carefully
+    # balanced to both work with ONNX export and XLA. In particular XLA
+    # prefers ints, cumsum defaults to output longs, and ONNX doesn't know
+    # how to handle the dtype kwarg in cumsum.
     mask = tensor.ne(padding_idx).int()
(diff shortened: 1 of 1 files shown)


fairseq/utils.py
```diff
@@ -172,8 +172,8 @@ def make_positions(tensor, padding_idx, onnx_trace=False):
 
     Position numbers begin at padding_idx+1. Padding symbols are ignored.
     """
-    mask = tensor.ne(padding_idx).long()
-    return torch.cumsum(mask, dim=1) * mask + padding_idx
+    mask = tensor.ne(padding_idx).int()
+    return (torch.cumsum(mask, dim=1) * mask).long() + padding_idx
```
myleott: This is causing unit tests to fail. This should work:
```
return (torch.cumsum(mask, dim=1, dtype=torch.int) * mask).long() + padding_idx
```

I'll update it and merge.

myleott: Unfortunately I'm not able to make changes to the PR, so you'll need to update (or I can create a new PR). Either way, please confirm that the above works for XLA.

taylanbil: Thanks Myle, I'm going to test this as soon as possible, however xla is broken atm due to https://github.com/pytorch/xla/issues/827, and it might take some time. I'll get back to you asap.

Issue #887, opened 2019-07-18
change "self.left_pad" from True to False after "src_tokens" is converted from left-pad to right-pad


vineetk1:
Lines 211 through 218 in [https://github.com/pytorch/fairseq/blob/master/fairseq/models/lstm.py](url) convert **src_tokens** from left-padded to right-padded. However, **self.left_pad** is not changed from True to False.
Suggestion is to add a new line between existing lines 218 and 219 within the if-block. This new line will have the following content:
self.left_pad = False

huihuifan:
Can you make a pull request?

vineetk1:
Pull Request submitted

myleott:
Left comments in #893 😄

PR #893, opened 2019-07-19
Pull Request for Issue #887

vineetk1:


myleott:
I don't think this is right. `self.left_pad` indicates that our inputs will have padding on the left. That tells the model it should convert the padding direction on every forward. If you reset `self.left_pad` then it will only happen on the first forward and future forward's will be broken :/

vineetk1:
I see your point based on how the original developers of this code are interpreting `self.left_pad.` Another way of interpreting `self.left_pad` is that it reflects the padding in `src_tokens`. If the padding for `src_tokens` is changed from left to right then `self.left_pad` should change from True to False. In the next forward, if `src_tokens` is left-padded then `self.left_pad` should be set to True.
I also note that `self.left_pad` is not needed anywhere else in this class. By the way, I am making changes to this code, including forward, for a Dialog application that I am adding to fairseq. Even though my extra code in forward does not rely on `self.left_pad` but if it did then it will find that  `self.left_pad` is True but `src_tokens` is incorrectly padded on the right.
I am not going to sweat over this issue but I will make two suggestions:
(1) Add a comment that explains the interpretation of `self.left_pad`.
(2) A better suggestion is to remove `self.left_pad` and use `src_tokens` to determine whether it is left-padded or right-padded.
I can do a PR on either one of these suggestions, or I am perfectly fine if you want to close this issue.

PR #904, opened 2019-07-24
added patience for early stopping

gvskalyan:
Added patience flag to support early stop when the validation metric does not improve for certain no of patience epochs


gvskalyan:
@lematt1991 please consider reviewing this

sai-prasanna:
@myleott Any update on this ?

ecly:
This would be a nice addition.

PR #917, opened 2019-07-29
Also ignore Vim/Emacs editor backup file suffixed with "~" while importing models

freewym:


myleott:
Sorry, that change was done in a hurry. Switching back to .py ending, but adding support for models in directories this way: https://github.com/pytorch/fairseq/pull/920

PR #948, opened 2019-08-01
Use ==/!= to compare str, bytes, and int literals

cclauss:
Identity is not the same thing as equality in Python.


myleott:
Thanks!

PR #965, opened 2019-08-03
RoBERTa: Fix BOS token in `encode`

julien-c:


myleott:
Sorry the format isn't super clear. Every sequence begins with a beginning-of-sentence (`<s>`) symbol. Every sentence ends with an end-of-sentence (`</s>`) and we use an extra end-of-sentence (`</s>`) as a separator.

Example (single sentence): `<s> a b c </s>`
Example (sentence pair): `<s> d e f </s> </s> 1 2 3 </s>`

So the current code should be correct.

The origin for this format is that we organized the training data as one sentence per line and and extra blank line between documents. We take chunks of tokens and prepend a single `<s>` at the beginning of each chunk.

julien-c:
I see. Thanks!

Issue #971, opened 2019-08-04
shlex.quote problem in scripts/spm_train.py


freewym:
At [this line](https://github.com/pytorch/fairseq/blob/master/scripts/spm_train.py#L17), `shlex.quote` will add a quote embracing the whole command line argument, e.g., if the command line is: 
`python spm_train.py --input=data/text --user_defined_symbols=[UNK]`

The arguments part will be parsed into `--input=data/text '--user_defined_symbols=[UNK]'` because of the appearance of square brackets, and it will lead to an error:

`sentencepiece_trainer.cc(116) LOG(INFO) Running command: --input=data/text '--user_defined_symbols=[UNK]'`
`Traceback (most recent call last):
  File "../../scripts/spm_train.py", line 17, in <module>
    spm.SentencePieceTrainer.Train(" ".join(map(shlex.quote, sys.argv[1:])))
OSError: Not found: unknown field name "'--user_defined_symbols" in TrainerSpec.`

where there is a single quote before `--user_defined_symbols`. How about removing the use of `shlex.quote` at [this line](https://github.com/pytorch/fairseq/blob/master/scripts/spm_train.py#L17)?

freewym:
@myleott Any comments on this?

PR #972, opened 2019-08-05
remove shlex.quote in scripts/spm_train.py

freewym:
to resolve the issue #971


myleott:
I guess this will require a bit more escaping on the user's side, but looks more flexible, thanks!

PR #995, opened 2019-08-07
Integrate with Apache Arrow/Plasma in-memory store for large datasets

myleott:
Datasets with many examples can generate very large indexes in TokenBlockDataset (and possibly elsewhere). When using `--num-workers>0` these indexes are pickled and transferred via a multiprocessing pipe, which is slow and can fail if the index grows beyond 4GB (~0.5B examples). Apache Arrow has an in-memory store called Plasma that will offload these arrays to shared memory, which both reduces duplication of the data and avoids needing to pickle.


myleott:
> looks good to me, can you check if those plasma_store processes get killed after training is interrupted?

They do! :)

PR #1015, opened 2019-08-13
Speedup asr_prep_json.py using ProcessPoolExecutor

cpuhrsch:
Creates a fixed number of Processes each computing an equal chunk instead of one thread per entry. Speedup on librispeech 960h is about 8x.

### DIFF
--- a/examples/speech_recognition/datasets/asr_prep_json.py
+++ b/examples/speech_recognition/datasets/asr_prep_json.py
@@ -77,7 +77,6 @@ def main():
         raise Exception('No labels found in ', args.labels_path)
 
     samples = []
-    i = 0
     for path, _, files in chain.from_iterable(os.walk(path) for path in args.audio_dirs):
         for f in files:
             if f.endswith(args.audio_format):
@@ -88,7 +87,6 @@ def main():
                 if utt_id not in labels:
                     continue
                 samples.append(Sample(os.path.join(path, f), utt_id))
-                i += 1
 
     utts = {}
     num_cpu = args.num_cpu
--- a/examples/speech_recognition/datasets/asr_prep_json.py
+++ b/examples/speech_recognition/datasets/asr_prep_json.py
@@ -56,7 +56,8 @@ def main():
                         help="aggregated input labels with format <ID LABEL> per line",
                         type=argparse.FileType('r', encoding='UTF-8'))
     parser.add_argument("--spm-model", required=True,
-                        help="sentencepiece model to use for encoding")
+                        help="sentencepiece model to use for encoding",
+                        type=argparse.FileType('r', encoding='UTF-8'))
     parser.add_argument("--dictionary", required=True,
                         help="file to load fairseq dictionary from",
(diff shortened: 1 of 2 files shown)


examples/speech_recognition/datasets/asr_prep_json.py
```diff
@@ -21,19 +20,32 @@
...
-def process_sample(aud_path, lable, utt_id, sp, tgt_dict):
-    input = {}
-    output = {}
-    si, ei = torchaudio.info(aud_path)
-    input["length_ms"] = int(si.length / si.channels / si.rate / MILLISECONDS_TO_SECONDS)
-    input["path"] = aud_path
+def process_samples(samples, labels, sp_path, tgt_dict):
+    utts = {}
+
+    import sentencepiece as spm
```
okhonko: nit: Seems like there is no way to create json w/o sentencepiece so we will have to import it anyway. Can we import it at the top of the module and fail fast in case if it's not found?

examples/speech_recognition/datasets/asr_prep_json.py
```diff
@@ -44,19 +56,17 @@ def main():
                         help="aggregated input labels with format <ID LABEL> per line",
                         type=argparse.FileType('r', encoding='UTF-8'))
     parser.add_argument("--spm-model", required=True,
-                        help="sentencepiece model to use for encoding",
-                        type=argparse.FileType('r', encoding='UTF-8'))
```
okhonko: nit: why not leave `type=argparse.FileType` and get check if file exists for free? `sp.Load(sp_path)` probably does check if file exists but we can get it earlier with argparse

examples/speech_recognition/datasets/asr_prep_json.py
```diff
@@ -66,23 +76,27 @@ def main():
...
                 if utt_id not in labels:
                     continue
                 samples.append(Sample(os.path.join(path, f), utt_id))
 
     utts = {}
-    num_cpu = multiprocessing.cpu_count()
-    with concurrent.futures.ThreadPoolExecutor(max_workers=num_cpu) as executor:
-        future_to_sample = {executor.submit(process_sample, s.aud_path, labels[s.utt_id], s.utt_id, sp, tgt_dict): s for s in samples}
-        for future in concurrent.futures.as_completed(future_to_sample):
+    num_cpu = args.num_cpu
```
okhonko: nit: assert num_cpu > 0

examples/speech_recognition/datasets/asr_prep_json.py
cpuhrsch: TextIOWrapper isn't serializable, so this can't be used.

examples/speech_recognition/datasets/asr_prep_json.py
cpuhrsch: This is at a per-process level, so I'm importing it here. We could probably do it at the top as well, but we don't need it in the rest of the script (I removed the above import).

PR #1083, opened 2019-08-28
Update to Classical Structured Prediction branch

cimeister:
I'm trying to replicate some of the results from Classical Structured Prediction Losses for Sequence to Sequence Learning. I've moved some of the code from the original branch into a branch of the latest fairseq repo. So far, I've only done the risk minimization criterion but would really like feedback if anyone has time! I've tried running it with experimental settings from the paper, but haven't been able to replicate results (command taken from the readme of the classic_seqlevel branch). BLEU score on the validation set of IWSLT’14 German-English seems to hang at around 32 after a few epochs (same as when I start fine tuning with risk minimization). Is there something I'm missing in my code?

On another note, would there be any interest from the community in me updating the other criterion from the classic_seqlevel branch to work with the current releases of pytorch/fairseq?


cmeister747:
Update: After ~20 epochs of training with risk minimization, I get improvements on par with the those published in the paper. Baseli

myleott:
Cool! We have been meaning to update the code and merge it into master, but never got around to rerunning all the experiments to reproduce the results.

We actually also had an effort to move some of the logic into a FairseqTask, rather than a Criterion, to keep everything more self-contained. That way you could just do `--task=translation_struct`. Would love it if you could look over that version and give any feedback (esp. as it compares to this one): https://github.com/pytorch/fairseq/compare/seq_task?expand=1.

cmeister747:
Ah yes, moving into it's own task would make a lot more sense. I'm new to Fairseq so still getting used to the layout! I'll checkout that branch and play around with it

myleott:
Thanks! Please feel free to push any changes to that branch (bugfixes, documentation, additional criterions, etc.), and if you do get a chance to reproduce the paper results using it, then it'd be great if you could submit a PR for that branch and we could merge it :)

cmeister747:
Just an update, so I've been running some experiments on the `seq_task` branch and things aren't going too well. I'm using the same pretrained model (convolutional architecture trained for ~250 epochs) as for the experiments on my branch (in this PR). But while average training cost is going down on my branch, it ends up increasing with the code from the seq_task branch. I haven't spent much time looking into why this is happening but if you have any intuition for the cause, it'd be great to know it ahead of time. I'll spend some time on it this week...

Commands I'm using:
on `seq_task` branch
`fairseq-train data-bin/iwslt14.tokenized.de-en -a fconv_iwslt_de_en --clip-norm 0.1 --momentum 0.9 --lr 0.25 --dropout 0.3 --max-tokens 500 --seq-max-len-a 1.5 --seq-max-len-b 5 --criterion sequence_risk --task translation_struct --seq-beam 16  --reset-optimizer`
on this branch
`fairseq-train data-bin/iwslt14.tokenized.de-en -a fconv_iwslt_de_en --clip-norm 0.1     --momentum 0.9 --lr 0.25    --dropout 0.3 --max-tokens 500     --seq-max-len-a 1.5 --seq-max-len-b 5 --criterion sequence_risk     --seq-combined-loss-alpha 0.5  --seq-beam 16  --reset-optimizer`

myleott:
Interesting, thanks for looking into this! At least one difference is probably this modification in the seq_task branch: https://github.com/pytorch/fairseq/commit/fb4dda92059bdaac2b403bfd50c43579198bd1c3.

I wonder if that should make such a big difference though...

cmeister747:
Sorry for the delay on this - didn't have time and then running experiments took a while. 

Here's what I'm seeing going on. First, model performance when trained with the scores exponentiated (the original version from the commit above) is definitely better than without. When the scores aren't exponentiated, validation cost (100 - avg Bleu) goes up slightly at each epoch and then plateaus around 70 (so Bleu of 30) after starting from ~64 (so Bleu of 36) but test set accuracy decreases quite a bit ( -.6 bleu after 10 epochs and - 2 bleu after 20 epochs). But when scores are exponentiated, validation cost decreases and test set accuracy is improved after 10 epochs. 

Isn't exponentiating the scores part of the formula normally in minimum risk training? I'm confused as to why it's removed in the above commit. From a mathematical perspective, the exponentiation would make the probabilities "peakier", which I would imagine is desired at this point in training. Just wondering, have you tried something like the deterministic annealing procedure the Smith/Eisner use [here](https://www.aclweb.org/anthology/P06-2101)? Sorry for all the questions!

Another strange thing: there's a big difference when I train using 1 gpu vs. 8. With the current version of the code, the model degrades quite a bit when I train with 1 gpu (i.e. cost increases steadily but test set bleu absolutely plummets (gets down to 16 after 20 epochs of training)). However, I get much more reasonable results when training with 8 gpus. Can you think of any reason why that might be? I was incredibly meticulous about making sure all other factors were the same.

In sum:

with scores exponentiated: test set bleu on model improves slightly, validation set cost goes down and plateaus

with score not exponentiated: test set bleu decreases, validation set cost increases and plateaus

with scores not exponentiated training on a single gpu: test set bleu goes to hell, validation set cost increases quite a bit and plateaus

all hyper parameters are either defaults or set as in the command in the comments above.

Would love to hear your thoughts!

cmeister747:
Ah, just re-read the paper and understand now that the exponentiation was taken out since that would make it done twice with the softmax.

PR #1097, opened 2019-08-30
Implementation of the WeCNLP abstract "Cross+Self-Attention for Transformer Models"

stephanpeitz:
This PR implements a new attention module which combines cross-attention (encoder-decoder attention) and the decoder self-attention. This work was accepted as an abstract at WeCNLP 2019 (https://www.wecnlp.ai/wecnlp-2019).

Cross+Self-Attention reduces the amount of parameter and increases the inference speed without any degradation in translation quality.
More details can be found in the attached [abstract](https://github.com/pytorch/fairseq/files/3561282/paper.pdf)

### DIFF
--- a/fairseq/models/transformer.py
+++ b/fairseq/models/transformer.py
@@ -122,6 +122,12 @@ class TransformerModel(FairseqEncoderDecoderModel):
                                  'Must be used with adaptive_loss criterion'),
         parser.add_argument('--adaptive-softmax-dropout', type=float, metavar='D',
                             help='sets adaptive softmax dropout for the tail projections')
+        parser.add_argument('--no-cross-attention', default=False, action='store_true',
+                            help='do not perform cross-attention')
+        parser.add_argument('--cross-self-attention', default=False, action='store_true',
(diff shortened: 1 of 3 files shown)


myleott:
Sorry for the delay in reviewing/merging this! This looks pretty straightforward, will aim to get it merged today.

fairseq/models/transformer.py
```diff
@@ -280,13 +291,24 @@ def reorder_encoder_out(self, encoder_out, new_order):
...
+                encoder_out['encoder_states'][idx] = state.index_select(1, new_order)
         return encoder_out
 
     def max_positions(self):
         """Maximum input length supported by the encoder."""
         if self.embed_positions is None:
             return self.max_source_positions
         return min(self.max_source_positions, self.embed_positions.max_positions())
+        
+    def buffered_future_mask(self, tensor):
```
myleott: Is this actually needed in the encoder? I can't see where it's used. There's already a definition in the decoder if that's sufficient?

PR #1130, opened 2019-09-12
Updated readme for multi-gpu instructions

ngoyal2707:


myleott:
Ah, let's prefer `--ddp-backend=no_c10d`. I actually found `--find-unused-parameters` doesn't work for the CommonsenseQA task for some reason.

myleott:
Fixed in #1140

PR #1169, opened 2019-09-24
Readme.md edit

udnet96:
Adding CNN model of summarization


udnet96:
Ah, got it. I thought all models which used fairseq are listed in this README.
Sorry for the confusion.

PR #1373, opened 2019-11-14
Ret self attn

Ueeek:


myleott:
Can you share more context about what this PR is trying to do?

Also it seems to have a merge conflict.

Issue #1315, opened 2019-10-28
FairseqMultimodel doesn't have forward_decoder


twelveth:
When I try to inference model, inheriting from `FairseqMultimodel` (I've created custom `MultilingualLSTMModel` with `MultilingualTransformerModel` as example) with either fairseq-generate or fairseq-interactive, it crashes, saying
```
AttributeError: 'MultilingualLSTMModel' object has no attribute 'forward_decoder'
```
There is no such method both in `FairseqMultiModel` and `BaseFairseqModel` (which is parent of `FairseqMultiModel`).

I suppose, the same issue will be with multilingual transformer, because both it and my class inherit from `FairseqMultiModel`.

trilobi:
+1

lematt1991:
Resolved by 4c5934ac61354d9b6d164f7317905e4ac2ae1064

PR #1379, opened 2019-11-15
New MultilingualLSTMModel and FairseqMultiModel fix

twelveth:
1. #1315 fix
2. Added new `MultilingualLSTMModel`, taking `MultilingualTransformerModel` as example


myleott:
This looks great, thank you!

Can you please add a test here: https://github.com/pytorch/fairseq/blob/master/tests/test_binaries.py

Issue #1403, opened 2019-11-20
Bug report: Finetuning of Structured Dropout transformer


mingruimingrui:
I'd like to open up a discussion to fix a bug I bumped into.

```
RuntimeError: Error(s) in loading state_dict for TransformerModel:
	Missing key(s) in state_dict: "decoder.layers.1.self_attn.k_proj.weight", ...
```

### Context

After training a transformer model with structured dropout.
Further finetuning can be done by doing layer pruning followed by continuation of training on the reduced model. 

```
# First train with structured dropout
fairseq-train \
  --arch transformer --decoder-layerdrop 0.2 \
  ...

# Finetune on reduced model
fairseq-train \
  --arch transformer \
  --restore-file LAST_CHECKPOINT \
  --encoder-layers-to-keep 0,1,2,3,4,5 \
  --decoder-layers-to-keep 1,3,5 \
  --reset-dataloader \
  --reset-lr-scheduler \
  --reset-meters \
  --reset-optimizer \
  ...
```

However the model runs into bug on loading of the checkpoint.
```
# This would cause the error above
load_ensemble_for_inference([checkpoint_file], task=task, model_arg_overrides=None)

# This would work
load_ensemble_for_inference(
    [checkpoint_file], task=task,
    model_arg_overrides={"decoder_layers_to_keep":"0,1,2"}
)
```

The problem can be easily solved by updating the model args after layer pruning to set
`encoder_layers_to_keep` and `decoder_layers_to_keep` to `None`.

huihuifan:
Thanks! @myleott left a comment in your pull request, could you look into that? Appreciate the fix!

mingruimingrui:
@huihuifan Thanks for the note! Big fan of all the work you've been pushing out!

huihuifan:
@mingruimingrui thanks for the kind words. Please open an issue or email me if you have any questions!

jiezhangGt:
Hello, do you find that LayerDrop could reduce the performance？
When I set the args ```--decoder-layerdrop 0.3 --encoder-layerdrop 0.2``` and trained an NMT model named ```M_LayerDrop```，then evaluate  M_LayerDrop use the args ```--encoder-layers-to-keep 0,1,2,3,4,5 --decoder-layers-to-keep 0,1,2,3,4,5```，I find that M_LayerDrop's bleu is 2 bleu less than the baseline model which trained without layerdrop。
The sincerity anticipates your reply， thanks！

PR #1406, opened 2019-11-20
Quick fix for Structured Dropout checkpointing

mingruimingrui:
Here's a quick fix for https://github.com/pytorch/fairseq/issues/1403.

To keep it short, this fix allows the user to checkpoint a translation model properly after applying layer pruning on a restored transformer file.

### DIFF
--- a/fairseq/models/fairseq_model.py
+++ b/fairseq/models/fairseq_model.py
@@ -68,14 +68,6 @@ class BaseFairseqModel(nn.Module):
         """
         self.upgrade_state_dict(state_dict)
         new_state_dict = prune_state_dict(state_dict, args)
-
-        # Since layers are now pruned, *_layers_to_keep are no longer needed.
-        # This is more of "It would make it work fix" rather than a proper fix.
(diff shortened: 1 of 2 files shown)


fairseq/models/fairseq_model.py
```diff
@@ -68,6 +68,14 @@ def load_state_dict(self, state_dict, strict=True, args=None):
...
         self.upgrade_state_dict(state_dict)
         new_state_dict = prune_state_dict(state_dict, args)
+
+        # Since layers are now pruned, *_layers_to_keep are no longer needed.
+        # This is more of "It would make it work fix" rather than a proper fix.
+        if "encoder_layers_to_keep" in vars(args):
+            args.encoder_layers_to_keep = None
+        if "decoder_layers_to_keep" in vars(args):
+            args.decoder_layers_to_keep = None
+
```
myleott: Please move this to the end of prune_state_dict instead: https://github.com/pytorch/fairseq/blame/master/fairseq/checkpoint_utils.py#L397

PR #1436, opened 2019-11-27
Fix typo in BART CNN/DM finetuning

Colanim:
Small typo in the inference script, in the README for finetuning BART on CNN/DM dataset.


myleott:
Thanks!

PR #1438, opened 2019-11-27
Fix shape issue of `return_all_hiddens` in roberta

HMJiangGatech:
By default `return_all_hiddens` is False, the shape of `features` will be BxTxC.
If use `return_all_hiddens`, the shape of `features` will be TxBxC.
See 
https://github.com/pytorch/fairseq/blob/9398a2829596393b73f5c5f1b99edf4c2d8f9316/fairseq/modules/transformer_sentence_encoder.py#L227

### DIFF
--- a/fairseq/models/roberta/model.py
+++ b/fairseq/models/roberta/model.py
@@ -360,7 +360,7 @@ class RobertaEncoder(FairseqDecoder):
         )
         features = inner_states[-1]
         if return_all_hiddens:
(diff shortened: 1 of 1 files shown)


myleott:
Thanks, this is a good catch!

I think it makes more sense to fix the API for `TransformerSentenceEncoder` so that it always returns `T x B x C`, since the transpose there isn't even needed. I'll push a change on top of this PR before merging

PR #1444, opened 2019-11-28
Make use of args.skip_invalid_size_inputs_valid_test in fairseq-interactive

Roxot:
`fairseq-interactive` currently ignores the `--skip_invalid_size_inputs_valid_test` parameter. I think this is the intended use of the parameter.


myleott:
Thanks!

PR #1457, opened 2019-12-04
[fix] ensure self.scale_window is an integer

donglixp:
The scale window condition is `elif (self._iter - self._last_overflow_iter) % self.scale_window == 0`, which requires self.scale_window is an integer. If self.scale_window is float, the result of `(self._iter - self._last_overflow_iter) % self.scale_window` will never be 0.


lematt1991:
Nice catch, thanks!

donglixp:
@lematt1991 The default value is `2**14 / args.distributed_world_size / args.update_freq[0]`. For example, 1024 GPUs are used in Roberta training. So the default scale_window would be 2**14, which seems quite small. I am wondering what is the common practice to choose scale_window? Thanks!

Issue #1393, opened 2019-11-19
Error with mutilingual model and fairseq-generate


feralvam:
Hi,
I got an error when generating with a trained multilingual model. I hope you can help me understand what went wrong and how to fix it. Some context: I'm basically trying to use the multilingual architecture as a multitask model to combine different datasets for a monolingual task (each task is a "language" pair).

The command used for training a many-to-one model (i.e. shared decoder) is:

```
CUDA_VISIBLE_DEVICES=1,2,3 fairseq-train "${data_dir}/bin" \
  --ddp-backend=no_c10d \
  --task multilingual_translation --lang-pairs orig-simp,complex-simp,long-simp \
  --arch multilingual_transformer \
  --share-decoders --share-decoder-input-output-embed \
  --encoder-embed-path "${glove}" --encoder-embed-dim 300 --encoder-ffn-embed-dim 300 \
  --decoder-embed-path "${glove}" --decoder-embed-dim 300 --decoder-ffn-embed-dim 300 \
  --encoder-attention-heads 5 --decoder-attention-heads 5 \
  --encoder-layers 4 --decoder-layers 4 \
  --optimizer adam --adam-betas '(0.9, 0.98)' \
  --lr 0.0005 --lr-scheduler inverse_sqrt --min-lr '1e-09' \
  --label-smoothing 0.1 --dropout 0.3 --weight-decay 0.0001 \
  --criterion label_smoothed_cross_entropy --max-update 10000 \
  --warmup-updates 4000 --warmup-init-lr '1e-07' \
  --max-tokens 4000 --update-freq 4 \
  --save-dir "${model_dir}" --tensorboard-logdir "${log_dir}" \
```

Training proceeds without problems. Now, I want to generate the output for the 'test' subset of one of the "language" pairs (orig-simp) that the model was trained on. 

```
fairseq-generate "${data_dir}/bin" \
  --path "${model_dir}/${checkpoint_name}.pt" \
  --lang-pairs orig-simp,complex-simp,long-simp \
  --task multilingual_translation --source-lang orig --target-lang simp \
  --batch-size 128 --beam 5 --remove-bpe=sentencepiece \
  --gen-subset test > "${experiment_dir}/outputs/${output_name}.out"
```

After running the command I get the following error:

```
/experiments/falva/tools/fairseq/fairseq/models/fairseq_model.py:280: UserWarning: FairseqModel is deprecated, please use FairseqEncoderDecoderModel or BaseFairseqModel instead
[...]

PR #1470, opened 2019-12-07
fix bug #1393

MrHuhoo:
add args in interactive.py of multilingual_translation task.
- sentencepiece_vocab : path to sentencepiece.bpe.model

### DIFF
--- a/fairseq/options.py
+++ b/fairseq/options.py
@@ -506,6 +506,10 @@ def add_generation_args(parser):
                        help='maximum iterations for iterative refinement.')
     group.add_argument('--iter-decode-force-max-iter', action='store_true',
                        help='if set, run exact the maximum number of iterations without early stop')
+    group.add_argument('--iter-decode-with-beam', default=1, type=int, metavar='N',
+                       help='if > 1, model will generate translations varying by the lengths.')
+    group.add_argument('--iter-decode-with-external-reranker', action='store_true',
+                       help='if set, the last checkpoint are assumed to be a reranker to rescore the translations'),
     group.add_argument('--retain-iter-history', action='store_true',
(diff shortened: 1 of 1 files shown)


fairseq/options.py
```diff
@@ -506,6 +506,10 @@ def add_generation_args(parser):
                        help='maximum iterations for iterative refinement.')
     group.add_argument('--iter-decode-force-max-iter', action='store_true',
                        help='if set, run exact the maximum number of iterations without early stop')
+    group.add_argument('--iter-decode-with-beam', default=1, type=int, metavar='N',
+                       help='if > 1, model will generate translations varying by the lengths.')
+    group.add_argument('--iter-decode-with-external-reranker', action='store_true',
+                       help='if set, the last checkpoint are assumed to be a reranker to rescore the translations'),
```
myleott: I think we can actually move all of these into the task and then when decoding you can do:
```generate.py --task transformer_lev (...)```. Lemme check with @kahne and see if we can clean this up.

File: fairseq/options.py [hunk 2]
```diff
@@ -522,6 +526,8 @@ def add_interactive_args(parser):
                        help='read this many sentences into a buffer before processing them')
     group.add_argument('--input', default='-', type=str, metavar='FILE',
                        help='file to read from; use - for stdin')
+    group.add_argument('--sentencepiece_vocab',require=True,
+                       help='path to sentencepiece.bpe.model')
```
myleott: This shouldn't be needed... does it work if you do:
```
fairseq-interactive (...) --bpe=sentencepiece --sentencepiece-vocab (...)
```

myleott:
Thanks for identifying this issue!

Issue #1370, opened 2019-11-13
No such file or directory: 'C\\dict.txt'


Alkabann:
When I try:

import torch
camembert = torch.hub.load('pytorch/fairseq', 'camembert.v0')
I have this error message:
No such file or directory: 'C\\dict.txt'

I have seen another post which seems to raise the same issue:
C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\tasks\masked_lm.py
in setup_task():
I CHANGED
paths = args.data.split(':') 
TO
paths = [args.data]

but iI have exactly the same error after this...

  File "<ipython-input-7-96fab3386019>", line 1, in <module>
    camembert = torch.hub.load('pytorch/fairseq', 'camembert.v0')

  File "C:\ProgramData\Anaconda3\lib\site-packages\torch\hub.py", line 359, in load
    model = entry(*args, **kwargs)

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\models\roberta\model.py", line 239, in from_pretrained
    **kwargs,

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\hub_utils.py", line 54, in from_pretrained
    arg_overrides=kwargs,

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\checkpoint_utils.py", line 182, in load_model_ensemble_and_task
    task = tasks.setup_task(args)

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\tasks\__init__.py", line 17, in setup_task
    return TASK_REGISTRY[args.task].setup_task(args, **kwargs)

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\tasks\masked_lm.py", line 69, in setup_task
    dictionary = Dictionary.load(os.path.join(paths[0], 'dict.txt'))

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\data\dictionary.py", line 188, in load
    d.add_from_file(f, ignore_utf_errors)

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\data\dictionary.py", line 205, in add_from_file
    raise fnfe

  File "C:\Users\aymer/.cache\torch\hub\pytorch_fairseq_master\fairseq\data\dictionary.py", line 199, in add_from_file
[...]

PR #1496, opened 2019-12-13
Possible fix for Masked_LM path calculation

egeozsoy:
At least under windows it wasn't working for me.

Possible fix for the issue described here https://github.com/pytorch/fairseq/issues/1370. Not sure if the change brakes other things though


fairseq/tasks/masked_lm.py
```diff
@@ -4,6 +4,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import os
+from pathlib import Path
```
myleott: No need to use pathlib, os.path.join works on Windows too (but see below)

fairseq/tasks/masked_lm.py
```diff
@@ -64,9 +65,7 @@ def __init__(self, args, dictionary):
 
     @classmethod
     def setup_task(cls, args, **kwargs):
-        paths = args.data.split(':')
-        assert len(paths) > 0
-        dictionary = Dictionary.load(os.path.join(paths[0], 'dict.txt'))
+        dictionary = Dictionary.load(str(Path(args.data) / 'dict.txt'))
```
myleott: The issue is a bit different. We wanted to support multiple `data` arguments, since when training over large datasets it's often easier to shard the data and iterate over each shard during training.

We used `':'` as the separator, since that character isn't allowed in Unix paths. Unfortunately it breaks for Windows users. The correct solution is to use `os.pathsep` instead of hardcoding `':'`.

fairseq/tasks/masked_lm.py
```diff
@@ -64,9 +65,7 @@ def __init__(self, args, dictionary):
 
     @classmethod
     def setup_task(cls, args, **kwargs):
-        paths = args.data.split(':')
```
myleott: Please just change this to:
```
paths = args.data.split(os.pathsep)
```

myleott:
I'll submit a new PR with the fix.

PR #1532, opened 2019-12-20
Return attention weights along other outputs in Transformer Encoder

### LINKED ISSUE #1484: Return attention in TransformerEncoderLayer
In Transformer **decoder**, attention is computed and returned, allowing user to access it later :

https://github.com/pytorch/fairseq/blob/828c1ca7522278aa6c1eaf91fe0425b8f40dd832/fairseq/modules/transformer_layer.py#L239

However in Transformer **encoder**, attention is ignored :

https://github.com/pytorch/fairseq/blob/828c1ca7522278aa6c1eaf91fe0425b8f40dd832/fairseq/modules/transformer_layer.py#L95

Therefore we cannot access it later.

---

**Is there any reason why encoder's attention is not returned ?**  
I think it would be possible to add `encoder_attn` to :

https://github.com/pytorch/fairseq/blob/4c0d432ec74e0737c5132cde18182ce3f94599c0/fairseq/models/transformer.py#L298-L303

astariul:

Fixes #1484.

### DIFF
--- a/fairseq/models/transformer.py
+++ b/fairseq/models/transformer.py
@@ -361,7 +361,7 @@ class TransformerEncoder(FairseqEncoder):
         x = F.dropout(x, p=self.dropout, training=self.training)
         return x, embed
 
-    def forward(self, src_tokens, src_lengths, cls_input=None, return_all_hiddens=False, **unused):
+    def forward(self, src_tokens, src_lengths, cls_input=None, return_all_hiddens=False, return_all_attn=False, **unused):
         """
         Args:
             src_tokens (LongTensor): tokens in the source language of shape
@@ -370,6 +370,8 @@ class TransformerEncoder(FairseqEncoder):
                 shape `(batch)`
             return_all_hiddens (bool, optional): also return all of the
                 intermediate hidden states (default: False).
+            return_all_attn (bool, optional): also return all of the
+                intermediate layers' attention weights (default: False).
 
         Returns:
             namedtuple:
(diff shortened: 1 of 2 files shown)


astariul:
Any idea from where is the error ?

fairseq/models/transformer.py
```diff
@@ -397,15 +403,18 @@ def forward(self, src_tokens, src_lengths, cls_input=None, return_all_hiddens=Fa
...
         encoder_states = [] if return_all_hiddens else None
+        encoder_attn = [] if return_all_attn else None
 
         # encoder layers
         for layer in self.layers:
             # add LayerDrop (see https://arxiv.org/abs/1909.11556 for description)
             dropout_probability = random.uniform(0, 1)
             if not self.training or (dropout_probability > self.encoder_layerdrop):
-                x = layer(x, encoder_padding_mask)
+                x, attn = layer(x, encoder_padding_mask, need_head_weights=return_all_attn)
```
myleott: The problem is that when `return_all_attn=False` then you only return `x`. You need to do something like:
```
layer_out = layer(x, encoder_padding_mask, need_head_weights=return_all_attn)
x = layer_out[0]
attn = layer_out[1] if len(layer_out) > 1 else None
```

astariul: I'm confused because `attn` is always returned from EncoderLayer :

https://github.com/pytorch/fairseq/blob/3311af162f5806defd519c70688200171c22b867/fairseq/modules/transformer_layer.py#L118

(I did it like this to follow the DecoderLayer)

---

Also from the tests :

>x, attn = layer(x, encoder_padding_mask, need_head_weights=return_all_attn)
>ValueError: too many values to unpack (expected 2)

Meaning more than 2 values are returned ??

I couldn't run the tests locally, I think I should focus on this before anything else...

Thanks for the help ! I will come back to this PR after understanding the tests better !

myleott: You can run the tests with `python setup.py test`

PR #1577, opened 2020-01-03
fix when the ensembled models do not have a decoder, e.g., FairseqEnc…

freewym:
…oderModel


myleott:
Looks good, thanks!

freewym:
@myleott Not sure why the checks failed...

PR #1594, opened 2020-01-06
replace symlinks with actual files in `fairseq_cli`

erip:
# Before submitting

## What does this PR do?

This PR replaces symlinks to files sitting in the project root with the files they're linked to in order to make the process of building from source cross platform. Currently `fairseq_cli` cannot be built from source on `master` because the symlinks use [unix paths](https://github.com/pytorch/fairseq/blob/master/fairseq_cli/eval_lm.py#L1).

This PR also inadvertantly fixes a bug of a missing symlink to [`fairseq-validate`](https://github.com/pytorch/fairseq/blob/master/setup.py#L167).

## PR review    


erip:
A side-effect of this is that the repo looks leaner, but that's just a preference thing I guess. 😄

erip:
@myleott a completely separate option is to create a `fairseq.cli` module which can then be invoked with `python -m` and this should work irrespective of how fairseq is built. This would still be different than how users might be doing things now, but is a nice option to consider. Wdyt? Happy to follow your recommendations if that makes things easier.

erip:
8573e5b addresses the request directly. I'm not sure if there's a way to only symlink on `pip install -e .`, but that might just be some seuptools magic with which I am not familiar.

I'd still be interested to discuss potentials for cleaning this up further, perhaps as a "breaking change" (not really, but you get the idea) that changes users interactions with the repo a tiny bit. @myleott totally your call - if you think that's a better change for later, we can merge this as is and clean this up further at a later point.

myleott:
> a completely separate option is to create a fairseq.cli module
> I'd still be interested to discuss potentials for cleaning this up further, perhaps as a "breaking change"

I think both of these are good suggestions, and might as well be implemented as a breaking change all at once. I'll circulate this internally among the other devs to see if anyone has strong opinions about it.

For now let's merge this to unblock the Windows build.

erip:
This has gotten very messy. Will create a separate PR...

PR #1629, opened 2020-01-17
Fix padding_idx logical error in Adaptive Input

Huffon:
# Before submitting

## What does this PR do?

I think if we keep pass **padding index of vocabulary** as `padding_idx` to adaptive embedding layers, 
there will be no chance to train some words.

e.g. If `cut_off` is (20000,60000) and vocab is larger than 60000, 
we can't learn[**20,000+padding_idx**]th word and [**60,000+padding_idx**]th word.
Because those words' ids will be **padding_idx** by subtraction logic and eventually get zero tensors.

So, I changed `self.padding_idx` to `None` after assign vocab's `padding_idx` 
**for the first time at head embedding representation**.


myleott:
Thanks, nice catch!

PR #1634, opened 2020-01-20
Fix logging of training sets (fixes #1632)

### LINKED ISSUE #1632: Training loss is mixed with valid loss
## 🐛 Bug

After the latest introduction of metrics, the training loss printed as log is different from before, which I suspect is because  its value is somewhat mixed with valid loss

### To Reproduce

Steps to reproduce the behavior (**always include the command you ran**):

You can reproduce it by observing the difference of training losses before/after setting `--disable-validation`

### Environment

 - fairseq Version (e.g., 1.0 or master): master
 - PyTorch Version (e.g., 1.0): 1.1
 - OS (e.g., Linux):
 - How you installed fairseq (`pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version:
 - GPU models and configuration:
 - Any other relevant information:

myleott:
* fix: mid-epoch validation metrics were previously polluting training metrics
* fix: mid-epoch metrics were not properly saved/restored in checkpoints
* added tests, both for metrics and for mid-epoch reproducibility

### DIFF
--- a/fairseq_cli/train.py
+++ b/fairseq_cli/train.py
@@ -141,6 +141,7 @@ def should_stop_early(args, valid_loss):
         return should_stop_early.num_runs > args.patience
 
 
+@metrics.aggregate('train')
 def train(args, trainer, task, epoch_itr):
     """Train the model for one epoch."""
     # Initialize data iterator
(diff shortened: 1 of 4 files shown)


fairseq_cli/train.py
```diff
@@ -174,17 +174,17 @@ def train(args, trainer, task, epoch_itr):
...
+            and num_updates > 0
+        ):
+            valid_losses = validate(args, trainer, task, epoch_itr, valid_subsets)
+            checkpoint_utils.save_checkpoint(args, trainer, epoch_itr, valid_losses[0])
+
+        if num_updates >= max_update:
+            break
 
     # log end-of-epoch stats
     stats = get_training_stats(agg.get_smoothed_values())
```
freewym: According to my test, This fix doesn't seems to completely fix the train loss issue.

agg is out of the scope of `with metrics.aggregate() as agg:`. Is it supposed to be like that?

myleott: The scope just controls which log messages are aggregated. The results can be accessed inside or outside of the scope via `agg`.

The problem before was that `validate` was called within the scope of the training `agg`, so mid-epoch valid metrics would get accumulated under the train one.

> According to my test, This fix doesn't seems to completely fix the train loss issue.

What is the test?

myleott: Yes, you're right, I need to update this.

PR #1689, opened 2020-02-09
[WIP] refactor registry to include factory capability

### LINKED ISSUE #1688: Consider use of prebuilt registry libraries
## 🚀 Feature Request

Fairseq should consider using prebuilt function/class registries internally

### Motivation

The current iteration of fairseq's internal architecture registry is slightly confusing. It ties itself very tightly to [argparse](https://github.com/pytorch/fairseq/blob/master/fairseq/registry.py#L17), has no mechanism for building with [non-`argparse.Namespace` arguments](https://github.com/pytorch/fairseq/blob/master/fairseq/registry.py#L41), and encourages a lot of [black magic](https://github.com/pytorch/fairseq/blob/42935dcc8c2d390547d14c29cc71f3f3f379bcfb/fairseq/registry.py#L65) with default arguments in the argument parser. It would be better if the registry logic could be punted to some separate library which is well-tested and more idiomatic.

### Pitch

Ideally the general architecture wouldn't change: there would be registries for `FairseqTask`, `Fairseq...Model`, `FairseqOptimizer`, `FairseqCriterion`, etc. What would likely change is that there would be no bifurcation of registration and instantiation of these classes. For example, under the paradigm of [class-registry](https://class-registry.readthedocs.io/en/latest/), you'd have something like...

```python
from class_registry import ClassRegistry

optimizer_registry = ClassRegistry()

@optimizer_registry.register("adafactor")
class FairseqAdafactor(FairseqOptimizer):
    def __init__(self, lr=1e-4, adafactor_eps=1e-3, clip_threshold=0.0, ...):
        ...

# Pass the non-defaults when looking up the class
opt = optimizer_registry.get("adafactor", lr=1e-5, clip_threshold=0.001)
```

This is highly desirable because it decouples completely from argument parsing, makes code more compositional by allowing `setup_task` to work strictly in terms of registries, and it requires less maintenance of code that isn't strictly "business logic" as far as fairseq is concerned. 

### Alternatives

There are two main options: leave the code as-is or explore registry libraries. The primary registry libraries that I know are [`class-registry`](https://class-registry.readthedocs.io/en/latest/) and [`catalogue`](https://github.com/explosion/catalogue). There may be others, but I assume they'll all be approximately equivalent.

### Additional Context

One question to explore before diving deeper is how this might interact with the `--user-dir` flag: can a user just import the appropriate registry from `fairseq` and register their own architectures? 

Another question is about how to communicate defaults to a user -- in the above formulation, defaults get handled by the class rather than by the user-facing command-line arg mechanism, so how can those defaults percolate back to a user? Are there global defaults? Do defaults get presented to a user currently? Is it a matter of documentation? Will we need black magic? ⚡️

erip:
# Before submitting

## What does this PR do?
Fixes #1688

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq/optim/__init__.py
+++ b/fairseq/optim/__init__.py
@@ -26,6 +26,7 @@ optimizer_registry = registry.Registry()
 
 build_optimizer, register_optimizer, OPTIMIZER_REGISTRY = registry.setup_registry(
     '--optimizer',
+    optimizer_registry,
     base_class=FairseqOptimizer,
     default='nag',
 )
--- a/fairseq/optim/__init__.py
+++ b/fairseq/optim/__init__.py
@@ -25,6 +25,7 @@ build_optimizer, register_optimizer, OPTIMIZER_REGISTRY = registry.setup_registr
     default='nag',
 )
 
+optimizer_registry = registry.Registry()
 
 # automatically import any Python files in the optim/ directory
 for file in os.listdir(os.path.dirname(__file__)):
--- a/fairseq/optim/__init__.py
+++ b/fairseq/optim/__init__.py
@@ -5,11 +5,15 @@
 
 import importlib
 import os
+import logging
 
 from fairseq import registry
 from fairseq.optim.fairseq_optimizer import FairseqOptimizer
 from fairseq.optim.fp16_optimizer import FP16Optimizer, MemoryEfficientFP16Optimizer
 from fairseq.optim.bmuf import FairseqBMUF  # noqa
+
+
+
 
 
 __all__ = [
@@ -18,6 +22,7 @@ __all__ = [
     'MemoryEfficientFP16Optimizer',
 ]
 
+optimizer_registry = registry.Registry()
 
 build_optimizer, register_optimizer, OPTIMIZER_REGISTRY = registry.setup_registry(
     '--optimizer',
@@ -25,7 +30,6 @@ build_optimizer, register_optimizer, OPTIMIZER_REGISTRY = registry.setup_registr
     default='nag',
 )
 
-optimizer_registry = registry.Registry()
 
 # automatically import any Python files in the optim/ directory
 for file in os.listdir(os.path.dirname(__file__)):
--- a/fairseq/criterions/__init__.py
+++ b/fairseq/criterions/__init__.py
@@ -9,9 +9,11 @@ import os
 from fairseq import registry
 from fairseq.criterions.fairseq_criterion import FairseqCriterion
 
+criterion_registry = registry.Registry()
 
 build_criterion, register_criterion, CRITERION_REGISTRY = registry.setup_registry(
     '--criterion',
+    criterion_registry,
     base_class=FairseqCriterion,
     default='cross_entropy',
 )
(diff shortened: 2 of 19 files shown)


myleott:
Unfortunately, I think we'll need to continue supporting Namespace's and the current registry API for backward compatibility, since we have a lot of internal users who register criterions/tasks/models using the existing API, so API changes can't be made so lightly. Maybe we can start from the desired interface for constructing these objects, and work backwards from there?

I'd like if we could support:
```
# named arguments (new API)
optimizer = fairseq.optim.Adadelta(params, lr=1e-4, eps=1e-6)

# builder with namespace (legacy API)
optimizer = fairseq.optim.build_optimizer(
    argparse.Namespace(optimizer='adadelta', lr=1e-4, adadelta_eps=1e-6),
    params
)

# builder with named arguments (new API)
optimizer = fairseq.optim.get_optimizer('adadelta')(params, lr=1e-4, eps=1e-6)
```

I'm not 100% sure we need to support the last API, but it should be easy enough to do and could be helpful.

I think we can get rid of `optimizer_config`, since it's only used within each Optimizer. Then something like this should work and not involve too many API changes:
```
@register_optimizer('adadelta')
class Adadelta(FairseqOptimizer):

    def __init__(self, params, lr, eps=1e-8):
        super().__init__()
        self._optimizer = torch.optim.Adadelta(
            params,
            lr=lr,
            eps=eps,
        )

    @classmethod
    def from_args(cls, args: argparse.Namespace, params):
        return cls(
            params,
            lr=args.lr,
            eps=args.adadelta_eps,
        )

    @staticmethod
    def add_args(parser):
        # remains unchanged
        # (...)
```

erip:
@myleott thanks for the quick review. I reverted the breaking changes and will try to refactor the legacy API to use the new API. I'll also roll in the classmethod to build from args, which I think will make things a little simpler. Would it be good to log a deprecation warning from the `register_x` and `build_x` methods of the legacy registry API, too? 

Also, is it safe to remove `optimizer_config`? The checkpointing logic is a total black box to me, so I don't know if removing that info from the optimizer would cause an incompatibility.

fairseq/optim/__init__.py
```diff
@@ -25,6 +25,7 @@
     default='nag',
 )
 
+optimizer_registry = registry.Registry()
```
ckczzj: ？

erip:
@myleott Do you have strong opinions about [this commit](https://github.com/pytorch/fairseq/pull/1689/commits/f9304079ec53c17c5f35d6ccb6704b50338aaf47#diff-cd8ba3594cee9320482106851c8428e8R58)? I think maybe this is a better path - by passing in the new registry object, we can defer calls to actually use that API.

myleott:
Passing in the Registry object looks fine.

I am questioning changing the recommending syntax from `register_optimizer` -> `optimizer_register.register` though. Since we can easily support the existing syntax (as you've done), `register_optimizer` seems a bit shorter to write, doesn't require us to change all the tutorials/docs to use new syntax, and reduces cognitive load for people familiar with the existing syntax. In terms of user-facing API, is there any benefit of `optimizer_register.register` over `register_optimizer`?

erip:
These are good points. :-) I guess even if the entire legacy API were scrapped, we could still even do `register_model = model_registry.register` without breaking things.

PR #1704, opened 2020-02-14
Scriptable simple sequence generator

liuchen9494:
Summary: Fairseq JIT scriptable Simple Sequence Generator. This scriptable Sequence Generator is fully compatiable with fairseq transformerModel meeting LATTE team demand. In the following Diff, it will support ensemble model for Sequence Generator. The test have verified this implementation is able to load previous sequence generator test, and given out same results.

Differential Revision: D19745260


gvskalyan:
Can you please point to tests, to use scriptable sequence generator

PR #1710, opened 2020-02-15
Enable freezing encoder weights in RoBERTa sentence prediction

prihoda:
# Before submitting

## What does this PR do?
Fixes https://github.com/pytorch/fairseq/issues/1711

### DIFF
--- a/fairseq/models/roberta/model.py
+++ b/fairseq/models/roberta/model.py
@@ -290,7 +290,7 @@ class RobertaEncoder(FairseqDecoder):
             weight=self.sentence_encoder.embed_tokens.weight,
         )
 
-    def forward(self, src_tokens, features_only=False, return_all_hiddens=False, masked_tokens=None, **unused):
+    def forward(self, src_tokens, features_only=False, return_all_hiddens=False, masked_tokens=None, freeze_encoder=False, **unused):
         """
         Args:
             src_tokens (LongTensor): input tokens of shape `(batch, src_len)`
@@ -299,6 +299,9 @@ class RobertaEncoder(FairseqDecoder):
                 `(batch, src_len, embed_dim)`.
             return_all_hiddens (bool, optional): also return all of the
                 intermediate hidden states (default: False).
+            masked_tokens (BoolTensor, optional): tokens masked during
+                masked LM training (default: None).
+            freeze_encoder (bool, optional): freeze encoder weights (default: False).
 
         Returns:
             tuple:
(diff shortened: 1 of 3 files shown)


myleott:
One downside of this approach is that it still uses dropout in the encoder, even though the weights are frozen. I think it'd be more natural to disable dropout in the encoder and keep dropout only in the classification head.

We actually did something similar recently, but opted to do it in the model instead. What do you think about adding a new `freeze_encoder` boolean option to `RobertaEncoder.forward`: https://github.com/pytorch/fairseq/blob/7d89d2e42995948aadf91922e51aaae7865ef0de/fairseq/models/roberta/model.py#L293-L313

It could do something like:
```python
with torch.set_grad_enabled(not freeze_encoder):
    if freeze_encoder:
        self.eval()  # disable dropout when encoder is frozen
    x, extra = self.extract_features(src_tokens, return_all_hiddens=return_all_hiddens)
    if freeze_encoder:
        self.train()
```

Then you can modify the criterion to pass in this option: https://github.com/pytorch/fairseq/blob/7d89d2e42995948aadf91922e51aaae7865ef0de/fairseq/criterions/sentence_prediction.py#L39-L43

prihoda:
@myleott Thanks for the comment. And what do you think about the possibility to cache the encoder output? Could this be done somehow using `lru_cache`? https://github.com/pytorch/fairseq/issues/1711#issuecomment-586591831

fairseq/models/roberta/model.py
```diff
@@ -307,7 +310,13 @@ def forward(self, src_tokens, features_only=False, return_all_hiddens=False, mas
                   is a list of hidden states. Note that the hidden
                   states have shape `(src_len, batch, vocab)`.
         """
-        x, extra = self.extract_features(src_tokens, return_all_hiddens=return_all_hiddens)
+        with torch.set_grad_enabled(not freeze_encoder):
+            was_training = self.training
+            if freeze_encoder:
+                self.eval()  # disable dropout when encoder is frozen
+            x, extra = self.extract_features(src_tokens, return_all_hiddens=return_all_hiddens)
+            if was_training and freeze_encoder:
```
prihoda: @myleott I added this check to make sure we are only enabling training if it has been enabled before

prihoda:
@myleott Now I guess I have to add `freeze_encoder` also to `TransformerEncoder` to support `sentence_prediction` in BERT. Can I just throw a `NotImplementedError` in case `freeze_encoder==True` or do I have to implement it there as well? Or shall I just add `**ununsed`?  https://github.com/pytorch/fairseq/blob/fdfdbec9e2f2eccfbdd331ae8d165d3ce05bcb73/fairseq/models/transformer.py#L420-L426

prihoda:
@myleott I rebased on master and added a test. Is there anything else I can do to make this mergeable?

PR #1714, opened 2020-02-16
[WIP] refactor namespaces

erip:
# Before submitting

- [x] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [x] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/master/CONTRIBUTING.md)?
- [x] Did you make sure to update the docs?   
- [x] Did you write any new necessary tests?  

## What does this PR do?
Fixes #1672 

## PR review    
Anyone in the community is free to review the PR once the tests have passed.     
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #1729, opened 2020-02-21
refactor namespaces in criterion interface

### LINKED ISSUE #1672: Replace use of argparse.Namespace with named args and kwargs where possible.
## 🚀 Feature Request

Remove `argparse.Namespace` parameter passing where possible.

### Motivation

Using `args` makes code unclear to read because it couples implementations with `argparse` tightly. Functions and classes should rely on what they need to function and no more. 

### Pitch

With the exception of `classmethod`s, arguments to functions and methods should be named and explicitly passed in. This makes the code:

- More readable -- no more cluttering the code with `args.X` in implementations that _only_ care about `X`
- More testable -- argparse is a runtime concern and won't/shouldn't be needed in tests
- More portable -- if I want to use a class from `fairseq` in some other code, I don't need to "mock" argument parsing logic.
- Less tightly coupled -- instead of guessing why some call-site passed in `args`, I know exactly what the minimal information required to compute some function will be.

### Alternatives

Leave the code as is. 😄 

### Additional context

As I begin looking at the refactor for #1648, I see `args` buried deep in the code. This refactor would be much easier if this were implemented first.

erip:
# Before submitting

## What does this PR do?
Fixes #1672 in part (part 1: [context](https://github.com/pytorch/fairseq/pull/1714#issuecomment-587507040))

## PR review    

## Did you have fun?


myleott:
Almost done with this one :) Had to make some changes to support backwards compatibility, since there's a lot of internal criterions that still use the old API.

erip:
Sorry for all the extra work, but I'm excited about these changes. Wiring up hydra should be a breeze after this.

myleott:
Sorry for the delay, still iterating on this. Will be merged soon...

erip:
Woo! Glad to see this land -- thanks, @myleott. I see that this was a lot of work -- if there's anything you want to push back to me for the other components, please don't hesitate.

myleott:
Yeah thanks for all your work getting this started!

The first one took a bit longer because it required some discussion and figuring out the right way to maintain backward compatibility. The key is providing a Legacy version of the base classes, for example the LegacyFairseqCriterion.

It’d be super helpful if you can add some of those for the other PRs, but in general I hope the next few of these should be easier to merge.

erip:
Awesome. Happy to do that - looks like a light lift after the hard decision making was made up front. 😄 Should I also make PRs to pytorch/translate in tandem (like the associated commit w/ this PR)? I was given the impression that translate and fairseq were going to merge eventually, but I don't suspect that changes the need to keep them in lock step.

MultiPath:
Hi, what is the purpose for this change which might cause all the user defined criterion broken?

erip:
The purpose of this change is to divorce components in fairseq from argparsing. If you want to use some model/criterion/task/etc. outside of fairseq, you basically cannot. We have retained a `FairseqLegacyCriterion` interface which is a minimal change to keep compatibility with the previous interface.

MultiPath:
I don't understand. 
The original criterion design already contained "add_args" functions to add specific arguments. I am not sure the current change for "from_args" is necessary and I felt it made code even confusing.

erip:
The original criterion design stored an `argparse.Namespace` as an instance variable which means that you need one to create a criterion which is inconvenient. Eventually all of these `{add,from}_args` methods will be replaced by proper configuration.

MultiPath:
Not sure what do you mean by "you need one to create a criterion which is inconvenient. "
I think the design of args has more freedom to add different configurations as needed.

erip:
Basically:

```
class Criterion:
    def __init__(self, args: argparse.Namespace):
        self.args = args # <-- this has nothing to do with what it means to be a criterion
```

MultiPath:
Ok.. If it is the case, everything needs to be changed in fairseq...

erip:
These changes are not as dramatic as they appear. Basically it's standardizing the code toward best practices which has a lot of benefit downstream in testing and consuming.

MultiPath:
I think using "args" makes it much easier to pass arguments which might be defined outside of criterions.

erip:
Why can't you pass it to the constructor directly?

MultiPath:
Also, if a function requires 10+ more arguments as inputs to "__init__", it will also be very difficult to read and work with.

erip:
Difficult to read: maybe. Consider abstracting something. 
Difficult to work with: I don't think I agree.

MultiPath:
Anyway, I still don't think it is a good idea to remove all "args" to build the current model/criterion/etc. 
There may be a lot of places where you need to have this "args" as input, otherwise it is easily to get 10+ or 20+ inputs in a complicated model. 
Just to be clear, since Myle has already accepted the PR, I just left my comments here.

erip:
To make this exceeding clear: if someone develops an interesting optimizer in fairseq that I want to use in my pytorch code, there's currently no way to do that unless I mock argument parsing. Optimizers should _not_ care about argument parsing much like models, lr schedulers, tokenizers, byte-pair encoders, loss functions, or just about any other component in software (outside of a main function) should not.

MultiPath:
..but this PR is about "loss functions".

erip:
I was giving an example. There are PRs for every other component, as well. See #1743, #1733, #1732, #1731, #1730

MultiPath:
Yes, I understand. 
I think it is **super** unnecessary and makes everything hard to read, change, and maintain our previous codebase. Especially for "tasks", "models".

myleott:
Sorry just catching up on this.

@MultiPath, I shared some context/motivation in the other PR (#1743), but I agree that we need to be careful here for several reasons:
1) Currently fairseq uses args everywhere, so we need to be careful not to break existing code. We introduced the "LegacyFairseqCriterion" base class that should be a drop-in replacement to make any existing code continue to work.
2) There are cases where using a single `args` is very helpful and we should continue to support this where it's appropriate.
3) The hope is that this will make it easier to use fairseq as a library. The motivation is to let people initialize and use fairseq components as standalone pieces in other projects, rather than only supporting the fairseq command-line usage.

myleott:
> The original criterion design already contained "add_args" functions to add specific arguments. I am not sure the current change for "from_args" is necessary

Hmm, seems this is a github bug. Please ignore the "Files Changed" shown in this PR. The actual commit is 719fb82db646d8b694e2a7174b800f94dcd65b6e, which included a number of changes. Notably, we didn't end up adding `from_args` and also introduced `LegacyFairseqCriterion`.

PR #1743, opened 2020-02-24
Refactor namespaces tasks

### LINKED ISSUE #1672: Replace use of argparse.Namespace with named args and kwargs where possible.

erip:
# Before submitting

## What does this PR do?
Fixes #1672 in part (part 6: [context](https://github.com/pytorch/fairseq/pull/1714#issuecomment-587507040))

## PR review    

## Did you have fun?


erip:
@myleott I think this is ready for import, too.

erip:
This will need to be reimported.

MultiPath:
I think it is very unnecessary and makes everything hard to read, change, and maintain our previous codebase, especially for "tasks".

erip:
Unless you're actively developing fairseq,`find $USER_DIR -type f -exec sed -i -e 's/FairseqTask/LegacyFairseqTask/g' {} \;` should take care of maintaining your code for the foreseeable future.

myleott:
Thanks @MultiPath, I agree this adds way too much boilerplate and makes the current task implementations challenging to read and modify. We will iterate on this.

The motivation for the refactor is to support using fairseq as a library and with other config systems (e.g., [facebookresearch/hydra](https://github.com/facebookresearch/hydra)). Currently this is difficult because `args` are passed around all over, and it's not clear what arguments are used by each component. One way of handling this is by using explicitly-named arguments, but there's a huge downside to this approach in the form of *boilerplate*, which is most evident with FairseqTask (this PR).

The original motivation for using argparse.Namespaces everywhere was to make it as fast as possible to add and experiment with new configuration options. This is crucial for research productivity, where one might add a dozen configuration options when developing/testing a new idea.

So, we need to strike a balance between these two motivations. I'll describe my thoughts and would love feedback.

1) I feel the explicitly-named arguments approach works well for things like criterions, optimizers, LR schedulers, etc., where the functionality is limited/self-contained and thus requires only a few named arguments. With our [automatic mapping between argparse.Namespace keys and keyword argument names](https://github.com/pytorch/fairseq/blob/43cf9c977b8470ec493cc32d248fdcd9a984a9f6/fairseq/criterions/fairseq_criterion.py#L31-L33) it require 3 lines of code to add a new argument: adding to `add_args`, and adding two lines to `__init__`.
2) For more complicated components like models and tasks, it's not uncommon for tasks to require 20+ named arguments. The current implementation in this PR doesn't have automatic mapping, so requires duplicating each argument 4 times: once in `add_args`, once in `setup_task`, twice in `__init__`. I have some thoughts on how to reduce this below.

### **Let's consider the number of lines of code needed to add a new argument to FairseqTask:**

#### In the current system: one line of code
```diff
def add_args(parser):
+    parser.add_argument('--my-new-argument')
```

#### Current PR: four lines of code
```diff
-def __init__(self):
+def __init__(self, my_new_argument):
+    self.my_new_argument = my_new_argument

def add_args(parser):
+    parser.add_argument('--my-new-argument')

def setup_task(cls, args):
-    return cls()
+    return cls(args.my_new_argument)
```

#### With [automatic mapping of arg names](https://github.com/pytorch/fairseq/blob/43cf9c977b8470ec493cc32d248fdcd9a984a9f6/fairseq/criterions/fairseq_criterion.py#L31-L33): three lines of code
```diff
-def __init__(self):
+def __init__(self, my_new_argument):
+    self.my_new_argument = my_new_argument

def add_args(parser):
+    parser.add_argument('--my-new-argument')

# setup_task will automatically map argparse.Namespace -> named arguments
```

#### If we automatically reconstruct `args` in `__init__`: two lines of code
```diff
-def __init__(self):
+def __init__(self, my_new_argument):
    self.args = self.convert_named_args_to_namespace()

def add_args(parser):
+    parser.add_argument('--my-new-argument')

# setup_task will automatically map argparse.Namespace -> named arguments

def convert_named_args_to_namespace(self):
    """Automatically populate any arguments used by __init__."""
    return argparse.Namespace(...)
```

What do people think about the last option here?

MultiPath:
> automatic mapping of arg names

I like the last option more. 
But does that mean it is inevitably adding all the arguments in the __init__ function every time we make a new task? For instance, the current translation task requires tons of arguments, and some of them will never be used and only work in many specific problems.

erip:
Personally I don't like `args` in the object at all. If the number of parameters of `__init__` is unmanageable or if some of the parameters only apply in some cases, this is begging for a better abstraction. I think this conversation is very useful, but I think thinking about A) commonalities of constructor patterns, B) possible abstractions for constructor arguments, and C) more refined tasks would all be very useful.

myleott:
Thanks for the feedback @MultiPath and @erip! I'll bring this up at the next fairseq meeting and will share any notes from the discussion back here.

Re: `args` in the object, I doubt we'll make a rigid declaration here either way -- there are tasks for which the number of args is small and it makes sense to enumerate in the `__init__` (e.g., audio_pretraining, cross_lingual_lm, language_modeling) and others where the added flexibility of passing around a single configuration object is helpful (e.g., when rapidly trying a new idea). I'd like to leave these decision to the Task author. Note that Hydra also includes support for OmegaConf objects that are very similar to `args`, so we'll probably want to support that where it makes sense.

Re: better abstractions, I totally agree. But that's beyond the scope of this PR. Hopefully this PR will highlight cases where better abstractions are required though (e.g., translation).

Re: "inevitably adding all the arguments in the init function", we will definitely provide a fallback option that does not require this.

myleott:
Another option:

#### Only things defined in `add_args` are present in `args`: one line of code
```diff
def __init__(self, args, distributed_world_size):
    # args only contains my_new_argument
    # any other arguments must be passed in explicitly

def add_args(parser):
+    parser.add_argument('--my-new-argument')
```

erip:
The problem with args is that there won't be any way to use hydra if we go that route. Hydra and argparse basically do not work together.

erip:
I think maybe if we go the route of typed configs, that might be a promising route. Something like:

```python

from abc import ABCMeta

class FairseqConfig(metaclass=ABCMeta):
    pass

class FairseqTask(metaclass=ABCMeta):
    def __init__(self, config: FairseqConfig)
        pass # intentionally not storing the config object
```

This gives you the structure that `argparse.Namespace` provides without coupling so tightly. Thoughts?

PR #1783, opened 2020-03-05
[#1782] Add max position params to speech recognition

### LINKED ISSUE #1782: Speech recognition can OOM with large audio sequences
## 🚀 Feature Request

Unlike other tasks such as `translation`, `speech_recognition` doesn't support the possibility to define max sequence lengths in order to filter examples not respecting them. This can lead to OOM during the training. So it would be nice to have the same feature in this task too.

### Motivation

Without it, several OOM can happen. This can cause other examples to be ignored and a slower training.

### Pitch

We propose to add the `--max-source-positions` and `--max-target-positions` args to the `speech_recognition` task.

### Alternatives

The only alternative I can think of is ignoring the OOMs...

mgaido91:
# Before submitting

## What does this PR do?
Fixes #1782.

## PR review    

## Did you have fun?

### DIFF
--- a/examples/speech_recognition/tasks/speech_recognition.py
+++ b/examples/speech_recognition/tasks/speech_recognition.py
@@ -77,6 +77,10 @@ class SpeechRecognitionTask(FairseqTask):
         parser.add_argument(
             "--silence-token", default="\u2581", help="token for silence (used by w2l)"
         )
+        parser.add_argument('--max-source-positions', default=2048, type=int, metavar='N',
+                            help='max number of tokens in the source sequence')
+        parser.add_argument('--max-target-positions', default=1024, type=int, metavar='N',
+                            help='max number of tokens in the target sequence')
 
     def __init__(self, args, tgt_dict):
         super().__init__(args)
@@ -132,3 +136,7 @@ class SpeechRecognitionTask(FairseqTask):
         """Return the source :class:`~fairseq.data.Dictionary` (if applicable
         for this task)."""
         return None
+
+    def max_positions(self):
+        """Return the max speech and sentence length allowed by the task."""
+        return (self.args.max_source_positions, self.args.max_target_positions)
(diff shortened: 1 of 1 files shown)


mgaido91:
the failure seems unrelated to this PR, but I don't know how to retrigger the tests, can anyone help me? Thanks.

erip:
@mgaido91 Windows never passes.

examples/speech_recognition/tasks/speech_recognition.py
```diff
@@ -77,8 +78,8 @@ def add_args(parser):
         parser.add_argument(
             "--silence-token", default="\u2581", help="token for silence (used by w2l)"
         )
-        parser.add_argument('--max-source-positions', default=2048, type=int, metavar='N',
-                            help='max number of tokens in the source sequence')
+        parser.add_argument('--max-source-positions', default=sys.maxsize, type=int, metavar='N',
+                            help='max number of frames in the source sequence')
         parser.add_argument('--max-target-positions', default=1024, type=int, metavar='N',
```
okhonko: We can make this default value as `sys.maxsize` as well, unless we know that there are no utterances in librispeech dataset with targets BPE units longer than 1024 .

mgaido91: Thanks for your reviews! I guess it depends on the BPE size used. In my case, I am using 8000 BPE and the longest target contains 132 tokens. 1024 is the default max for the translation task too. I can change it to `sys.maxsize` if you prefer, but IMHO this is reasonable. Thanks.

okhonko: Thanks for checking!

mgaido91:
Thanks for the review @okhonko . Is there anything else I can do to push this PR forward? Thanks in advance for the guidance.

PR #1798, opened 2020-03-08
fix epoch check for curriculum learning

freewym:
# Before submitting

- Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/master/CONTRIBUTING.md)?
- Did you write any new necessary tests?  

## What does this PR do?
Fixes epoch check for curriculum learning. `epoch_itr.epoch` has not been incremented before calling `next_epoch_itr()`. So the check for curriculum learning should look like the proposed fix.

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq_cli/train.py
+++ b/fairseq_cli/train.py
@@ -148,7 +148,7 @@ def train(args, trainer, task, epoch_itr):
     # Initialize data iterator
     itr = epoch_itr.next_epoch_itr(
         fix_batches_to_gpus=args.fix_batches_to_gpus,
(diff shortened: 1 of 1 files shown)


freewym:
@myleott I just realized it may still not solve the problem, if `args.curriculum==1`. It's difficult to differentiate between the start of the training and the end of epoch 1 where `epoch_itr.epoch == 1` in both cases. since now the initial epoch is 1 rather than 0.

myleott:
Yes, exactly. I'll patch your fix to handle this.

PR #1896, opened 2020-03-23
quant noise code, readme, start of adding quantization

huihuifan:
FUNCTIONALITY:
This diff provides two core pieces of functionality
- Adds training with quantization noise from "Training with Quantization Noise for Extreme Model Compression" - controlled by the "quant_noise" and "quant_noise_block_size" parameters. Added in embeddings, attention, FFN for BERT and Transformer LM training
- Adds quantization with product quantization based on code from "And the bit goes down: Revisiting the quantization of neural networks" (Stock et al, 2019). This is applied to a fairseq trained model to quantize after training. 

TODO:
-> @Pierre, look at quantization code
-> int4 and int8 quantization will be added soon. 

EVALUATED TEST CASES:

0. Training of LM and BERT models starts from scratch with no errors -> yes 

1. Retrain LM from scratch with code, no quantization, reproduces Wikitext-103 LM results -> yes, see /checkpoint/angelafan/qn_open_source_noise

2. Reload previously trained LM from scratch, not trained with quant noise, reproduces Wikitext-103 LM results -> yes 

3. Train LM from scratch with code, no trained with quant noise, reproduces Wikitext-103 LM results -> yes, see /checkpoint/angelafan/qn_open_source_baseline

4. Train BERT model from scratch with code, no quantization, training curve looks the same as before -> yes

5. Check wps during training and wps during inference, no large change from before -> yes

6. Check structured dropout isn't being applied at eval time -> yes 

7. Works in combination with LayerDrop -> yes


TopGun666:
Thanks for your great work. I wonder when the code of Quant-Noise could be accessed.

huihuifan:
code here: https://github.com/pytorch/fairseq/tree/master/examples/quant_noise?fbclid=IwAR3QWt-tD57HiQQRhyLTwhqdid7iuZpAgD8E_YQ4aqLhgbMUirjDLKgRp_I

jahutwb:
Hi!
I'm trying to quantize iwslt14 model from translation example,
https://github.com/pytorch/fairseq/tree/master/examples/translation
but facing some problems.
I train model with quant-noise by adding --quant-noise-pq argument,
`--quant-noise-pq 0.1 --quant-noise-pq-block-size 8`
but it doesn't train at all. Returns something like "and and and and and [...]" or ", , , , , , [...]" and so on.
I tried with block size 2, but the results was similar.

I tried to quantize model by training without quant-noise, but with 
`    --quantization-config-path examples/quant_noise/transformer_quantization_config.yaml` (I didn't change the yaml I found here)
to see if i can quantize anything.
It found 3 layers to quantize, because it wanted max_epoch argument to be divisible by 3. It trained quite well but didn't quantize anything.
`2020-06-08 09:54:59 | INFO | fairseq.quantization_utils | Non-compressed model size: 150.56 MB. After quantizing 0 layers, size                                      (indexing + centroids + other): 0.00 MB + 0.00 MB + 150.56 MB = 150.56 MB, compression ratio: 1.00x`

What am I doing wrong?

jahutwb:
I found that get_layers in quantization/pq/utils.py returns empty list
I print all layers and checked that regex.
I changed  "\\" to "\"   in config.yaml like that decoder\.layers\.\d+\.fc[12], and made sure that all_layers is a list of strings
` list(filter(r.match, all_layers))`
end it worked.

But still have problem with training with quant noise

PR #1897, opened 2020-03-24
Add quality-latency curve for the text-to-text baseline

jmp84:
Rendering can be checked here: https://github.com/jmp84/fairseq/blob/simulastsharedtask/examples/simultaneous_translation/docs/baseline.md


jmp84:
@xutaima here is the matplotlib code just in case you want to reuse for the speech curve:

import matplotlib
import matplotlib.pyplot as plt
import numpy as np

# Data for plotting
al = [-0.99, 1.13, 2.53, 2.91, 3.12, 4.21, 5.30, 5.88, 6.53, 7.16, 12.37, 19.85]
bleus = [7.33, 12.15, 13.29, 14.17, 15.15, 14.80, 15.99, 17.28, 17.03, 17.64, 19.53, 23.04]

fig, ax = plt.subplots()
ax.plot(al, bleus)

ax.set(xlabel='AL', ylabel='BLEU',
       title='BLEU vs AL, text-to-text wait-k baseline')
ax.grid()

fig.savefig('/tmp/waitk_txt_bleu_vs_al.eps', format='eps', dpi=1000)
fig.savefig('/tmp/waitk_txt_bleu_vs_al.png', format='png', dpi=1000)
plt.show()

PR #1951, opened 2020-03-31
examples/translation: Specifiy the correct bpe when loading WMT19

xhlulu:
# Description

In [examples/translation](https://github.com/pytorch/fairseq/tree/master/examples/translation), the code will not run if you change the model from `transformer.wmt16` to `transformer.wmt19`, since the BPE they are using are different. I corrected that with a note at the end of the section.

### DIFF
--- a/examples/translation/README.md
+++ b/examples/translation/README.md
@@ -65,6 +65,13 @@ zh2en.translate('你好 世界')
 # 'Hello World'
 ```
 
+If you are using a `transformer.wmt19` model, you will need to set the `bpe` argument to `'fastbpe'`:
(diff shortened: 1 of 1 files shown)


myleott:
Thanks for the PR! The tokenizer/BPE doesn't actually need to be specified ([it's already defined in the code](https://github.com/pytorch/fairseq/blob/master/fairseq/models/transformer.py#L74-L86)), so let's just remove those kwargs from the README.

xhlulu:
@myleott I updated my branch based on your recommendations; I removed the kwargs from the readme.

PR #1996, opened 2020-04-11
fix the bug when the output dir does not exist

xutaima:
This pull request fixed the bug for server when the output dir does not exist. When the --output is not none, the server will try to create the directory, skip writing predictions if failed. 

Also change the writing pattern to $output.text to $output/text.

### DIFF
--- a/examples/simultaneous_translation/eval/scorers/scorer.py
+++ b/examples/simultaneous_translation/eval/scorers/scorer.py
@@ -4,16 +4,18 @@ from vizseq.scorers.meteor import METEORScorer
 from examples.simultaneous_translation.utils.eval_latency import LatencyScorer
 from collections import defaultdict
 import json
+import os
 
 DEFAULT_EOS = '</s>'
 class SimulScorer(object):
     def __init__(self, args):
         self.tokenizer = args.tokenizer
+        self.output_dir = args.output
         if args.output is not None:
             self.output_files = {
-                "text": args.output + ".text",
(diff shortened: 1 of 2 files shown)


examples/simultaneous_translation/eval/scorers/scorer.py
```diff
@@ -4,16 +4,18 @@
...
         self.tokenizer = args.tokenizer
+        self.output_dir = args.output
         if args.output is not None:
             self.output_files = {
-                "text": args.output + ".text",
-                "delay": args.output + ".delay",
-                "scores": args.output + ".scores"
+                "text": os.path.join(args.output, "text"),
+                "delay": os.path.join(args.output,"delay"),
+                "scores": os.path.join(args.output,  "scores")
```
jmp84, line 18: one space instead of 2

examples/simultaneous_translation/eval/scorers/scorer.py
```diff
@@ -4,16 +4,18 @@
 from examples.simultaneous_translation.utils.eval_latency import LatencyScorer
 from collections import defaultdict
 import json
+import os
 
 DEFAULT_EOS = '</s>'
 class SimulScorer(object):
     def __init__(self, args):
         self.tokenizer = args.tokenizer
+        self.output_dir = args.output
```
jmp84, line 13: In start-server.sh, we have `--output $result_dir/eval` so it's not a directory.

PR #2040, opened 2020-04-21
add local evaluation for simultaneous translation

xutaima:
Usage:

- Text Translation

``` shell
python $fairseq/example/simultaneous_translation/eval/evaluate_local.py \
    --scorer-type text \
    --agent-type simul_trans_text \
    --tokenizer 13a \
    --src-file $src \
    --tgt-file $tgt \
    --data-bin $data_bin \
    --model-path $model \
    --src-splitter-type $src_splitter_type \
    --src-splitter-path $src_splitter_path \
    --tgt-splitter-type $tgt_splitter_type \
    --tgt-splitter-path $tgt_splitter_path \
    --num-threads 10 \
    --output
```
- Speech Translation

``` shell
python $fairseq/example/simultaneous_translation/eval/evaluate_local.py \
    --scorer-type speech \
    --agent-type simul_trans_speech \
    --tokenizer 13a \
    --tgt-file $json_file \
    --data-bin $data_bin \
    --model-path $model \
    --src-splitter-type $src_splitter_type \
    --src-splitter-path $src_splitter_path \
    --tgt-splitter-type $tgt_splitter_type \
    --tgt-splitter-path $tgt_splitter_path \
    --num-threads 10 \
    --output
```

### DIFF
--- a/examples/simultaneous_translation/eval/agents/agent.py
+++ b/examples/simultaneous_translation/eval/agents/agent.py
@@ -1,11 +1,14 @@
 from . import GET, SEND, DEFAULT_EOS
 import time
 from multiprocessing.pool import ThreadPool as Pool
-from functools import partial 
+from functools import partial
+
+
 class Agent(object):
     "an agent needs to follow this pattern"
     def __init__(self, *args, **kwargs):
         pass
+
     def init_states(self, *args, **kwargs):
         raise NotImplementedError
 
@@ -20,7 +23,7 @@ class Agent(object):
 
     def reset(self):
         raise NotImplementedError
-        
+
     def decode(self, session, low=0, high=100000, num_thread=10):
         corpus_info = session.corpus_info()
         high = min(corpus_info["num_sentences"] - 1, high)
--- a/examples/simultaneous_translation/scripts/start-local-eval.sh
+++ b/examples/simultaneous_translation/scripts/start-local-eval.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+set -e
+config=$1
+. $1
+
+model=$2
+
+if [ $scorer_type = "text" ]
+then
+    num_test=$(wc -l $src | cut -d " " -f1)
+else
+    num_test=$(grep length_ms $tgt | wc -l)
+fi
+echo $num_test
+
+chunk_size=300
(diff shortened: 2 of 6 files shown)


examples/simultaneous_translation/scripts/start-local-eval.sh
```diff
@@ -0,0 +1,44 @@
...
+echo Config: $(realpath $config)
+echo Model: $(realpath $model)
+echo Source: $src
+echo Target: $tgt
+echo Tokenizer: $tokenizer
+echo Agent: $agent_type
+
+user_dir=$(dirname "$0")/..
+
+echo Evaluatation starts at $(date +%Y/%m/%d-%H:%M:%S) &&
```
mjdousti, line 27: Do you want to calculate the runtime? If so, use `time` command instead.

examples/simultaneous_translation/eval/agents/agent.py
```diff
@@ -1,11 +1,14 @@
 from . import GET, SEND, DEFAULT_EOS
 import time
 from multiprocessing.pool import ThreadPool as Pool
-from functools import partial 
+from functools import partial
```
jmp84, line 4: How did it work before :O ?

examples/simultaneous_translation/scripts/start-local-eval.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -e
```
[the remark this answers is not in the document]
jmp84: Can you explain the advantages?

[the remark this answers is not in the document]
mjdousti: In `cmd1 | cmd2` line, if `cmd1` fails, the script won't fail as the return value would be taken from `cmd2`. `set -o pipefail
` ensures the script would fail when `cmd1` fails.

For a more detailed discussion, see https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

examples/simultaneous_translation/eval/agents/agent.py
mjdousti: These two lines are identical to me!

PR #2090, opened 2020-05-04
[#2022] Pass all net_inputs in SequenceGenerator

### LINKED ISSUE #2022: Recent SequenceGenerator update makes it unusable with sub-classes that have a different `net_input`
## 🐛 Bug

This is not a real bug, I rather want to discuss a recent change in master. 

**Situation**: My (transformer) encoder sub-class needs additional input during the forward path. Currently I'm feeding this to my encoder by adding it to the `net_input` dictionary. During validation/generation I need the additional input as well. This worked out of the box for generation, because the generator only used to forward the `net_input` to the encoder:

https://github.com/pytorch/fairseq/blob/630701eaa750efda4f7aeb1a6d693eb5e690cab1/fairseq/sequence_generator.py#L133

With a recent change of the `SequenceGenerator` this is not possible anymore. The generator now only feeds `src_tokens` and `src_lengths` to the encoder (probably for torchscript):

https://github.com/pytorch/fairseq/blob/57526c63433c0b1c997fc91c0881867532567266/fairseq/sequence_generator.py#L197-L200

This change requires me to sub-class the generator and add the input there as well. 

### Expected behavior

I should be able to add additional input to my encoder, without having to sub-class the generator. 

### Environment

 - fairseq Version (e.g., 1.0 or master): master

mgaido91:
# Before submitting

## What does this PR do?
Fixes #2022.

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq/sequence_generator.py
+++ b/fairseq/sequence_generator.py
@@ -194,10 +194,7 @@ class SequenceGenerator(nn.Module):
             self.min_len <= max_len
         ), "min_len cannot be larger than max_len, please adjust these!"
         # compute the encoder output for each beam
-        encoder_outs = self.model.forward_encoder(
-            src_tokens=encoder_input["src_tokens"],
-            src_lengths=encoder_input["src_lengths"],
-        )
+        encoder_outs = self.model.forward_encoder(encoder_input)
 
         # placeholder of indices for bsz * beam_size to hold tokens and accumulative scores
         new_order = torch.arange(bsz).view(-1, 1).repeat(1, beam_size).view(-1)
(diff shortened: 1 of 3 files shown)


myleott:
I think the issue was making it work with torchscript, but lemme "import" this so that the internal tests run and see if it works...

mgaido91:
I see your point, but this seems a too big regression to me, basically any model needing additional input won't be usable anymore to generate outputs... We can maybe consider having two versions of the `forward_encoder` method, one which does not pass additional inputs and can be used with torchscript and the other which does the opposite and switch among them either relying on a new flag on the input models or adding a new arg, so user can decide what to do. What do you think?

myleott:
I figured out how to both fix this and support torchscript. Fix incoming.

mgaido91:
sounds great, thanks. I am closing this then. Thanks.

myleott:
@mgaido91, landed in 11345a7608a6b73feb53046f6c0eef3dc4d3fc08. Btw, thanks for adding the test!

mgaido91:
thank you for the fix @myleott !

Issue #1785, opened 2020-03-06
DenoisingDataset EOS token not set properly for MultilingualDenoisingTask when --add-lang-token is set


dkavaler:
## 🐛 Bug

The [collater function](https://github.com/pytorch/fairseq/blob/master/fairseq/data/denoising_dataset.py#L355) for `DenoisingDataset` does not set the correct EOS token when used in the MultilingualDenoisingTask.

This is caught in [this assertion](https://github.com/pytorch/fairseq/blob/master/fairseq/data/data_utils.py#L41).

I believe this is because the EOS token [is set during the construction of the DenoisingDataset](https://github.com/pytorch/fairseq/blob/master/fairseq/tasks/multilingual_denoising.py#L158), but the call in the collater function uses `self.vocab.eos()`. This issue is propagated down to the [call to `data_utils.collate_tokens`](https://github.com/pytorch/fairseq/blob/master/fairseq/data/denoising_dataset.py#L28) when the `key` has the value of `'target'`.

One solution may be to also pass the `eos_idx` of the `target` language to the `collate` function. I am not sure if that affects the intended design.

### To Reproduce

Run `fairseq-train` using `--task multilingual_denoising` and `--add-lang-token`, with data preprocessed in a similar way as [cross lingual language model training](https://github.com/pytorch/fairseq/tree/master/examples/cross_lingual_language_model).

### Expected behavior

The `eos_idx` variable should be set to the correct value in the call to `collate`.

### Environment

 - fairseq Version (e.g., 1.0 or master): master
 - PyTorch Version (e.g., 1.0): 1.4.0
 - OS (e.g., Linux): Linux
 - How you installed fairseq (`pip`, source): source
 - Build command you used (if compiling from source): `pip install -e .` (in fairseq directory)
 - Python version: 3.7.4
 - CUDA/cuDNN version: N/A
 - GPU models and configuration: N/A
 - Any other relevant information:

### Additional context

I am attempting to train a multilingual denoising model from scratch. I could not find documentation for this in the repository (besides [this](https://github.com/pytorch/fairseq/blob/master/examples/mbart/README.md), which does not contain information on how to train from scratch).

villmow:
Once you have a command working for pretraining a multilingual denoising model could you post it here? It seems like they don't want to share the pretraining procedure of BART and mBART somehow (see  #1504 , #1614 , #1389 , ...)

dkavaler:
[...]

PR #2129, opened 2020-05-13
Fix EOS token in multilingual_denoising

cbaziotis:
Fixes https://github.com/pytorch/fairseq/issues/1785, where in multilingual_denoising the generic eos token is used instead of the `langid` token. 

Specifically, the problem is that the [collater](https://github.com/pytorch/fairseq/blob/master/fairseq/data/denoising_dataset.py#L355) in the denoising_dataset uses the generic `self.vocab.eos()` instead of `self.eos`, which in the case of `multilingual_denoising` is [set](https://github.com/pytorch/fairseq/blob/master/fairseq/tasks/multilingual_denoising.py#L159) to the language id of a given language.

See https://github.com/pytorch/fairseq/issues/1785#issuecomment-626171602 for an example of the issue.

### DIFF
--- a/fairseq/data/data_utils.py
+++ b/fairseq/data/data_utils.py
@@ -38,7 +38,12 @@ def collate_tokens(values, pad_idx, eos_idx=None, left_pad=False, move_eos_to_be
     def copy_tensor(src, dst):
         assert dst.numel() == src.numel()
         if move_eos_to_beginning:
-            dst[0] = eos_idx
+            # if no eos_idx is specified, then use the last token in src
+            if eos_idx is None:
+                dst[0] = src[-1]
+            # else use the given eos_idx
+            else:
+                dst[0] = eos_idx
             dst[1:] = src[:-1]
         else:
             dst.copy_(src)
--- a/fairseq/data/denoising_dataset.py
+++ b/fairseq/data/denoising_dataset.py
@@ -26,7 +26,9 @@ def collate(
     def merge(key, left_pad, move_eos_to_beginning=False):
         return data_utils.collate_tokens(
             [s[key] for s in samples],
-            pad_idx, eos_idx, left_pad, move_eos_to_beginning,
+            pad_idx=vocab.pad(),
+            eos_idx=None,  # use eos_idx of each sample instead of vocab.eos()
+            left_pad=left_pad, move_eos_to_beginning=move_eos_to_beginning,
         )
 
     id = torch.LongTensor([s['id'] for s in samples])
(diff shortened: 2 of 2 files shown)


cbaziotis:
I realized that even this won't work, because each batch is a mixture of many languages, and the way that denoising_dataset works, it uses the [same](https://github.com/pytorch/fairseq/blob/9a718e29855713a51877237b2dcc25e39c234c82/fairseq/data/denoising_dataset.py#L29)  `eos_idx` for all samples in the batch.

```bash
> print(list("".join([self.task.dictionary.symbols[t] for t in s]) for s in sample['net_input']['prev_output_tokens'][:5].tolist()))

['[de]<s>▁Das▁ist▁phänomenal,▁da▁kann▁man▁sich▁schon▁ausmalen,▁was▁in▁Leipzig▁los▁ist,▁wenn▁wir▁Erstliga-Fußball▁anbieten.</s>',
 '[de]<s>▁Offering▁something▁for▁everyone,▁the▁hemlines▁varied▁from▁just▁above▁booties▁to▁tapered▁at▁the▁ankle▁or▁flowing.</s>',
 '[de]<s>▁Construction▁will▁start▁in▁the▁second▁quarter▁of▁this▁year▁and▁the▁first▁windmills▁from▁Vestas▁will▁be▁delivered▁and▁installed▁in▁2018.</s>',
 '[de]<s>▁You\'ve▁probably▁heard▁the▁term▁"glycemic▁index"▁(GI)▁touted▁on▁more▁than▁a▁few▁diet▁commercials.</s>',
 '[de]<s>▁The▁retired▁footballer,▁whose▁been▁pictured▁enjoying▁quality▁time▁shopping▁with▁his▁brood▁in▁London▁earlier▁this▁week,▁was▁praised▁by▁London▁Ambulance▁Service.</s>']

```

cbaziotis:
So the problem was in [merge](https://github.com/pytorch/fairseq/blob/master/fairseq/data/denoising_dataset.py#L29) inside the collate function. It uses a single eos_idx which when `move_eos_to_beginning=True` and the `--add-lang-token` flag is used in the `multilingual_denoising` task, makes all the samples in a batch have the same langid token in the target sequence. This is problematic because each batch might contain a mixture of many languages. 

I changed [data_utils.collate_tokens](https://github.com/pytorch/fairseq/pull/2129/commits/b9fef8991a016575747a76b58c6a28983e294a93#diff-47dbda9fcad726ba8d5aa806c9af04ee) to set `dst[0] = src[-1]` when `eos_idx=None` and `move_eos_to_beginning=True` and passed eos_idx=None to it. When not using `--add-lang-token` in `multilingual_denoising`, the shared eos token will be used, which is set [here](https://github.com/pytorch/fairseq/blob/master/fairseq/tasks/multilingual_denoising.py#L147).

Issue #1778, opened 2020-03-05
Do you plan to add Google's T5 model?


tomekd:
## 🚀 Feature Request
[T5](https://arxiv.org/abs/1910.10683) model is one of the most promising s2s pre-trained models now. Do you have any plans to add it to fairseq? 

### Motivation

T5 achieves the best scores on the [SuperGLUE](https://super.gluebenchmark.com/leaderboard) benchmark, so it would be nice to have it.

### Alternatives

I have started implementing this basing on the huggingface implementation. It is  not hard, but I would like to avoid potential duplication of work.

myleott:
I recently added a huggingface/transformers submodule with the hope of adding wrappers around the various models implemented there (XLNet, ALBERT, T5), but only added GPT-2 at first: https://github.com/pytorch/fairseq/blob/master/fairseq/models/huggingface/hf_gpt2.py

We’d love a PR adding T5 :) I guess you would just need to add the model wrapper and modify the existing denoising task.

PR #2172, opened 2020-05-22
Implement T5 model

tomekd:
Solved  #1778
# Before submitting

## What does this PR do?
This MR provides a T5 model into fairseq. I implemented only the finetuning task (without masking one). I tested it on the trivia-qa task and got comparable results.

## PR review    

## Did you have fun?
Sure, I learned a lot about this model, especially finding bugs.


tomekd:
@myleott : any idea how to fix windows build?

myleott:
Ah, ignore the windows build, it always fails.

Thank you for adding this btw! We have someone testing it out now and we'll get back to you with feedback. One early piece of feedback is that it'd be really great if we could support the largest model sizes with model parallelism. Do you know if this is possible currently with the Huggingface implementation?

tomekd:
No idea. To be honest, I tested on fp32 only, because of the different implementation of fp16 (bfloat16) on TPU.
I never tried the biggest model since a single V100 with 32gb is too small and had no experience with PyTorch in distributed training.

tomekd:
@myleott : any updates?

fairseq/models/t5/t5_modules.py
```diff
@@ -0,0 +1,516 @@
...
+        norm_x = self.layer_norm(hidden_states)
+        y = self.DenseReluDense(norm_x)
+        layer_output = hidden_states + self.dropout(y)
+        return layer_output
+
+
+class T5Attention(nn.Module):
+    NEW_ID = itertools.count()
+
+    def __init__(self, config, has_relative_attention_bias=False, is_cross_attention=False):
```
pengwa, line 70: Hello, this file is based on https://github.com/huggingface/transformers/blob/master/src/transformers/modeling_t5.py (most like of commit 
https://github.com/huggingface/transformers/commit/bbf26c4e619cf42106163e1e2cd5ff98b936ff93#diff-7370db3a19209bf984cc40925aaf2b71 as I binary searched).

While I found there a few modifications, by intention. This is_cross_attention is one of them. If is_cross_attention is True (For  EncDecAttention), relative_attention_bias will use nn.Functional.embedding(rp_bucket, torch.zeros(...)) 
 instead of nn.Embedding(). The problem I observed here is: in the former one, the embedding dict is not trainable, they are all zero. I checked the  https://applica-public.s3-eu-west-1.amazonaws.com/fairseq-t5/t5-small.pt, which did not contain the relative_attention_bias for encdecattentions. But I do find the weight on latest huggingface https://s3.amazonaws.com/models.huggingface.co/bert/t5-small-pytorch_model.bin. 



1. 

)

tomekd: That's a great question. The huggingface checkpoint is created in the following way: initialize the whole model randomly and then overwrite weights from Google checkpoint. Thus, the huggingface checkopoint contains EncDec relative embeddings (that are not in Google one). I can provide a simple script I used to convert from Tensorflow to Fairseq.

Lastly, I started working on the unsupervised task (I should have it tested until the end of this month). In addition, I discovered that I did not implement keeping decoder state during decoding, so it is a bit slow if the generated text is long.

pengwa: Thanks @tomekd. I assume the script converting from TF ckpt to Fairseq also contains the necessary optimizer states besides the weights themselves, right ? If you can share with that, that would be helpful. Thanks a lot!

tomekd: Sure thing! I will do it later today.

tomekd: Here is the script I used to convert: [script](https://gist.github.com/tomekd/899adcfae511fb3a2175488e812dbd32) Let me in case of any problem. Once I converted the models, I did not use it anymore later.

pengwa: Thanks you so much!

weiqi1028:
Hi, is someone still actively working on this one? We would also like a T5 implementation in fairseq.

Issue #1274, opened 2019-10-19
Using ResamplingDataset causes trouble with filtering to large samples


villmow:
Hi, thanks for the awesome library! 

**Problem**: I use the ResamplingDataset to downsample frequent examples in a custom task. After every epoch a subset of the full training data is selected for the next epoch. This selection affects the filtering by size of the samples which is implemented in `get_batch_iterator`. However after the filtered indices are calculated, `set_epoch` is called again (with an incremented epoch count)! This causes a resampling, which again invalidates the filtered indices! 

**What happens**:
1. I reset the `dataset_to_epoch_iter` to get a new iterator after each epoch.
2. `FairseqTask.get_batch_iterator` is called
3. Inside  `get_batch_iterator` the datasets `ResamplingDataset.set_epoch` is called, in which the resampling is implemented:
https://github.com/pytorch/fairseq/blob/b8d024e9b8c20058dd7282f1418ebef00bfb8974/fairseq/tasks/fairseq_task.py#L135-L136
4. The samples are filtered by size
https://github.com/pytorch/fairseq/blob/b8d024e9b8c20058dd7282f1418ebef00bfb8974/fairseq/tasks/fairseq_task.py#L138-L152
5. An EpochIterator is instantiated for which `next_epoch_itr` is called
https://github.com/pytorch/fairseq/blob/b8d024e9b8c20058dd7282f1418ebef00bfb8974/fairseq/tasks/fairseq_task.py#L155-L164
6. The trainer calls `next_epoch_itr` and inside of it the epoch counter is incremented and then `ResamplingDataset.set_epoch` is called again, which resamples the data (and thus the filtering indices are wrong).
https://github.com/pytorch/fairseq/blob/b8d024e9b8c20058dd7282f1418ebef00bfb8974/fairseq/data/iterators.py#L187-L191
7. This iterator is used and samples with wrong sizes are filtered.

I hope I could describe the issue understandable. As a quick fix for me I uncommented the `set_epoch` call in EpochBatchIterator, which I guess is not the right approach.

What is the correct way to handle the change of sizes after each epoch?

cbaziotis:
I am training a model with the `multilingual_denoising` task, which uses the ResamplingDataset. Training crashes deterministically form me after the first epoch, which I think is because of this issue.

What happens is the after the first epoch, I get OOM exceptions and I have found that the reason is the `ntokens` in batches is random and usually larger than `args.max_tokens`.

myleott:
[...]

PR #2176, opened 2020-05-23
Fixes for various minor bugs

cbaziotis:
This PR fixes various small bugs, which are self-explanatory from each commit's message. Most of them are related to `multilingual denoising` and `semisupervised_translation` tasks. I also added a workaround for this issue https://github.com/pytorch/fairseq/issues/1274.

### DIFF
--- a/fairseq/data/denoising_dataset.py
+++ b/fairseq/data/denoising_dataset.py
@@ -26,7 +26,9 @@ def collate(
     def merge(key, left_pad, move_eos_to_beginning=False):
         return data_utils.collate_tokens(
             [s[key] for s in samples],
-            pad_idx, eos_idx, left_pad, move_eos_to_beginning,
+            pad_idx=vocab.pad(),
+            eos_idx=None,  # use eos_idx of each sample instead of vocab.eos()
+            left_pad=left_pad, move_eos_to_beginning=move_eos_to_beginning,
         )
 
     id = torch.LongTensor([s['id'] for s in samples])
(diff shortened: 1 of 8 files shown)


fairseq/tasks/multilingual_translation.py
```diff
@@ -74,9 +74,9 @@ def add_args(parser):
                             help='source language (only needed for inference)')
         parser.add_argument('-t', '--target-lang', default=None, metavar='TARGET',
                             help='target language (only needed for inference)')
-        parser.add_argument('--left-pad-source', default='True', type=str, metavar='BOOL',
+        parser.add_argument('--left-pad-source', default=True, type=str, metavar='BOOL',
                             help='pad the source on the left (default: True)')
-        parser.add_argument('--left-pad-target', default='False', type=str, metavar='BOOL',
+        parser.add_argument('--left-pad-target', default=False, type=str, metavar='BOOL',
```
myleott, line 79: Can you clarify why this change is needed? I believe we cast them in lines 122-123 below (similar to `tasks/translation.py`)

fairseq/data/denoising_dataset.py
```diff
@@ -26,7 +26,9 @@ def collate(
     def merge(key, left_pad, move_eos_to_beginning=False):
         return data_utils.collate_tokens(
             [s[key] for s in samples],
-            pad_idx, eos_idx, left_pad, move_eos_to_beginning,
+            pad_idx=vocab.pad(),
```
myleott, line 29: I think this change from `pad_idx` to `vocab.pad()` may not be needed, since the function is called from DenoisingDataset.collater with `pad_idx=self.vocab.pad`, right?

PR #2190, opened 2020-05-27
Add bart.base to README (fixes #2189)

### LINKED ISSUE #2189: Will there be a pre-trained BART-base model available?
Since there is a bart_base_architecture method, will there be an actual pre-trained 'base' version of BART available?

myleott:


chris4540:
Hi,

I would like to ask what is the pretraining objective to this model.

I cannot find the pretraining objective in the file and the paper.

It will be grateful if you can answer my question.

Best,
Chris Lin

Issue #2237, opened 2020-06-12
Refactor LSTM model


mingruimingrui:
## 🚀 Feature Request
update LSTM Encoder/Decoder module to be more compatible with the rest of the package.

### Motivation
Transformers and convs are great! But their power mostly comes in the form of the encoders.

A suggestion would be to mix and match encoder and decoder arch. Eg. Dynamic conv encoder + RNN decoder. That way the speed and const memory nature of RNN decoder can be used in conjunction with the performance of the conv encoder.

myleott:
This would be nice, PRs welcome!

mingruimingrui:
Here's a brief update. I'm still running a few experiments to figure out the best way to do this (prioritizing performance to speed tradeoff).

Experiments are on wmt14-en-de comparing untokenized BLEU. (Tokenized BLEU (without segmenting compound words) for baseline is 28.6 which is about the results reported by Vaswani et al.)
The encoder uses the same settings (6 layers, normalize before etc...) throughout.

## Architecture comparison
![image](https://user-images.githubusercontent.com/18568364/85695067-4695cd80-b70a-11ea-9b7f-f8366f7dcb1d.png)

## Decoder layer comparison
![image](https://user-images.githubusercontent.com/18568364/85695077-48f82780-b70a-11ea-87ad-8a35b4ffebf3.png)

PR #2272, opened 2020-06-25
WIP: Feature/lstm decode

mingruimingrui:
# Before submitting

## What does this PR do?
Implements transformer models with LSTM decoder. #2237
Over the standard transformer, these models would perform sequence generation with constant memory and linear time complexity against the length of sequences generated.

## Did you have fun?
Yes. I only wish I have more time to work on community initiatives and open source. Been too busy with work recently.

## Overview of new features
2 implementations of Transformer with LSTM decoder
- `LightLSTMDecodeTransformerModel`
- `LSTMDecodeTransformerModel`

## `LightLSTMDecodeTransformerModel`
The decoder is a rewritten version of `LSTMDecoder`, local testing suggests that context from the previous time step improves results substantially.
Of all transformer models with single-layer decoder, this model performs the best.
An interesting observation is that the performance of this model does not appear to variate with decoder depth.

## `LSTMDecodeTransformerModel`
The decoder is architecturally similar to the `TransformerDecoder`. The only difference is that self-attention blocks are replaced by LSTM layers.
In local testing, this model basically achieves the same performance as the standard transformer while being a little bit faster.

## Bugfixes
- Fixed docs for command-line tools

## Self-debates
- I've thought about simply refactoring the `LSTMDecoder` but realized that the current implementation highly depends on the `LSTMEncoder`. Disconnecting this dependency changes some fundamental logic. To avoid the rabbit hole of backward compatibility, whole new models are written from scratch.
- Why not just have a single LSTM decode model? Both of them are good in different scenarios though.

### DIFF
--- a/docs/command_line_tools.rst
+++ b/docs/command_line_tools.rst
@@ -17,7 +17,7 @@ Fairseq provides several command-line tools for training and evaluating models:
 
 fairseq-preprocess
 ~~~~~~~~~~~~~~~~~~
-.. automodule:: preprocess
+.. automodule:: fairseq_cli.preprocess
 
     .. argparse::
         :module: fairseq.options
@@ -29,7 +29,7 @@ fairseq-preprocess
 
 fairseq-train
 ~~~~~~~~~~~~~
-.. automodule:: train
+.. automodule:: fairseq_cli.train
 
     .. argparse::
         :module: fairseq.options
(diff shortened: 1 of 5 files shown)


mingruimingrui:
## Comparison of different arch tested
image

## Comparison models with different number of decoder layers
image

mingruimingrui:
I see that there's the current effort to convert sequence generation to a pure torch script implementation. Will wait for next fairseq release before asking to merge.

PR #2288, opened 2020-06-30
Fix UnicodeEncodeError with fairseq-generate (#2287)

### LINKED ISSUE #2287: UnicodeEncodeError with fairseq-generate on Devanagari characters
## 🐛 Bug

A `UnicodeEncodeError` is raised when trying to translate source text with Devanagari (presumably any Unicode) text with `fairseq-generate`. The error is raised [this line](https://github.com/pytorch/fairseq/blob/894ae64858b62927d849c0fbc05e8f55d680a4f1/fairseq_cli/generate.py#L33)  of fairseq_cli/generate.py: `with open(output_path, 'w', buffering=1) as h:`. It can be fixed by just adding a `encoding = 'utf-8'` to the `open` call. The PR #460 addressed this issue for a bunch of other call sites of `open`, but not this one (perhaps it didn't exist at the time). That PR also mentions inconsistencies between platforms, therefore I think this is just an issue on Windows. I'm happy to make a PR if this is the right fix.

### To Reproduce

1. Run "fairseq-generate ..." with a source language that contains Devanagari characters

### Expected behavior

No error is raised.

### Environment

 - fairseq Version: master
 - PyTorch Version: 1.4.0
 - OS: Windows
 - How you installed fairseq : `pip install git+https://github.com/pytorch/fairseq`
 - Python version: 3.6.8
 - CUDA/cuDNN version: 10.2

TovlyDeutsch:
# Before submitting

- [x] Did you make sure to update the docs?  (n/a)
- [x] Did you write any new necessary tests?  (n/a)

## What does this PR do?
Fixes #2287 where a file `open` without encoding would cause a `UnicodeEncodeError` when using `fairseq-generate` with Unicode characters on Windows. A similar change was made in #460 but this line seems to have been added after that PR.

## Notes for similar future fixes
The repo has some other calls to `open` with non-binary files (i.e. mode is not `rb` or `wb`) that may cause similar issues. Perhaps there could be a codemod to inject encoding into all of these call sites, unless there are situations in this repo where text files should have a different encoding? Alternatively or in addition, perhaps a linter rule/plugin could be added to check for this case like this one: https://pypi.org/project/flake8-file-encoding/.

### DIFF
--- a/fairseq_cli/generate.py
+++ b/fairseq_cli/generate.py
@@ -30,7 +30,7 @@ def main(args):
     if args.results_path is not None:
         os.makedirs(args.results_path, exist_ok=True)
         output_path = os.path.join(args.results_path, 'generate-{}.txt'.format(args.gen_subset))
-        with open(output_path, 'w', buffering=1) as h:
+        with open(output_path, 'w', buffering=1, encoding='utf-8') as h:
             return _main(args, h)
     else:
         return _main(args, sys.stdout)


myleott:
Merged in ffecb4e3496379edf5ecae1483df5b7e0886c264.

PR #2356, opened 2020-07-20
Fix fairseq-generate score printing (fixes #2355)

### LINKED ISSUE #2355: fairseq-generate not printing score
https://github.com/pytorch/fairseq/blob/93f5128509278f425afb6bcf0da574c0af0e0c16/fairseq/bleu.py#L58

I think **.format()** should be added to the return line as the latest sacrebleu.corpus_bleu() now returns an object not a string

FadyEssam:
I think .format() should be added to the return line as the latest sacrebleu.corpus_bleu() now returns an object not a string

### DIFF
--- a/fairseq/bleu.py
+++ b/fairseq/bleu.py
@@ -55,7 +55,7 @@ class SacrebleuScorer(object):
     def result_string(self, order=4):
         if order != 4:
             raise NotImplementedError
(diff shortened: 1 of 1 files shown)


DhruvaBansal00:
Any updates on this?

Issue #2351, opened 2020-07-20
Loading model parallel Megatron weights for fine-tuning


rakeshchada:
#### What is your question?
Hello. Is there a way to load weights of a model that was trained in model parallel fashion? (eg:- loading Megatron weights that are provided [here](https://github.com/pytorch/fairseq/tree/master/examples/megatron_11b)). The weights are split across 8 parts. Hence, the regular way of providing "--restore_file" argument doesn't work. The use-case here is to fine-tune the model.

#### What have you tried?
I've tried modifying the checkpoint loading code a bit to load each part's weights separately on each GPU but that creates a different exception as now some of the parameter shapes start to differ. The original Megatron-LM does provide [a script to merge partitions](https://github.com/NVIDIA/Megatron-LM/blob/master/tools/merge_mp_partitions.py) to a single model file. But this merging might make sense to do for evaluation but not for fine-tuning.

So, any pointers on how to load megatron model parallel weights for fine-tuning are appreciated. Thank you.

ngoyal2707:
the `--restore-file` should still work as we add suffix here: https://github.com/pytorch/fairseq/blob/master/fairseq/checkpoint_utils.py#L126

I might have only tested it for continued pretraining (and not finetuning, although they are almost same thing). can you please try naming the checkpoint as `checkpoint_last-model_part-x.pt`? Let me know how it goes.

rakeshchada:
@ngoyal2707 That doesn't seem to work either. I get the same error:
```
-- Process 7 terminated with the following error:
Traceback (most recent call last):
  File "/home/ubuntu/fairseq/fairseq/trainer.py", line 257, in load_checkpoint
    state["model"], strict=True, args=self.args
  File "/home/ubuntu/fairseq/fairseq/models/fairseq_model.py", line 93, in load_state_dict
    return super().load_state_dict(new_state_dict, strict)
  File "/home/ubuntu/anaconda3/envs/pytorch_p36/lib/python3.6/site-packages/torch/nn/modules/module.py", line 830, in load_state_dict
    self.__class__.__name__, "\n\t".join(error_msgs)))
RuntimeError: Error(s) in loading state_dict for ModelParallelTransformerLanguageModel:
        size mismatch for decoder.embed_tokens.weight: copying a param with shape torch.Size([6400, 3072]) from checkpoint, the shape in current model istorch.Size([29345, 3072]).
[...]

PR #2365, opened 2020-07-23
Fixes checkpoint_path while loading a model-parallel checkpoint

rakeshchada:
Fixes https://github.com/pytorch/fairseq/issues/2351


rakeshchada:
@ngoyal2707 Created this PR as we discussed in the above issue.

Issue #1536, opened 2019-12-21
Add lexical constraints in decoding


erip:
### Motivation

Guiding decoding toward certain outputs in seq2seq tasks is particularly desireable in some use cases (e.g., standardizing translations or text rewriting); this feature is currently unavailable in fairseq.

### Pitch

The ability to provide an optional glossary which indicates a desireable target given a source token to `fairseq-generate` which would then be used in decoding would fill this gap. 

### Alternatives

N/A

### Additional context

See [this paper](https://arxiv.org/abs/1804.06609) for the implementation details; Sockeye currently implements this algorithm [docs](https://awslabs.github.io/sockeye/inference.html#lexical-constraints). h/t @mjpost

erip:
This might be a better issue for `pytorch_translate`, but thought I might include it here. Let me know if it should move!

hadyelsahar:
+1 for this, I have previously Implemented using Faireseq a hacky version of Matt Post's constraint decoding above. I will be willing to help implementing this. 

Some questions:
- What would be the best way to provide those constraints. in sockeye they provide them tab separated in the input file . Should we keep that?

- shall i keep asking questions here if needed?

erip:
@hadyelsahar I think asking questions here is super reasonable. My 2cc is that a global constraints file (e.g., a glossary/termbase) would be a great way to provide the constraints as some simple file type -- TSV? CSV?

The way I envision this happening is...

A user has source and target files for train, validation, and test. `fairseq-preprocess` will train BPE -- once BPE has been trained, this could be applied to a user's provided  constraints file (and should happen if `--lexical-constraints` is provided). The train, valid, test, dicts, and constraints would then appear in the `dest_dir`. 

@myleott @mjpost does this seem reasonable?

hadyelsahar:
> My 2cc is that a global constraints file (e.g., a glossary/termbase) would be a great way to provide the constraints as some simple file type -- TSV? CSV?

For simplification I opt more to fwd the prob processing the termbase to the user and ask the user to directly provide list of phrases to appear for each test example in a tsv file. Like this we can handle more scenarios than MT where there isn't  a direct word alignment between inputs and outputs. LM & Img captioning for example.  

Given that we will handle only "phrases to appear" type constraints. I am happy with the rest of the proposal.

myleott:
Merged in bd1b35d9b7cb21b2e7c17201d831c17560265b67

PR #2402, opened 2020-07-31
Added constrained decoding (#1536)

mjpost:
# Before submitting

## What does this PR do?

This PR implements constrained decoding ([Hokamp & Liu, 2017](https://www.aclweb.org/anthology/P17-1141/); [Post & Vilar, 2018](https://www.aclweb.org/anthology/N18-1119/)) with vectorization for batching ([Hu et al., 2019](https://www.aclweb.org/anthology/N19-1090/)). In addition, it add *ordered constraints*, where the constraints are generated on the target side in order, with zero or more unconstrained tokens in between. This variant allows for optimizations that increase speed and BLEU scores (when testing with random scraps from the references).

### Usage and quick start

It works with `fairseq-interactive` via a new command-line option: `fairseq-interactive --constraints [ordered,unordered]`, defaulting to `ordered` if nothing is provided. When active, it will split lines from STDIN on `\t`, with separate constraints each separated by a tab. For example (after downloading the [Fairseq WMT19 German--English model](https://github.com/pytorch/fairseq/blob/master/examples/wmt19/README.md)):

```bash
echo -e "Die maschinelle Übersetzung ist schwer zu kontrollieren.\thard\tinfluence" \
  | [normalize.py](https://gist.github.com/mjpost/4c54446b7030d7c64b57461d27090650) \
  | [tok.py](https://gist.github.com/mjpost/ed7456f6a987c533102fc121678ed302) \
  | PYTHONPATH=$HOME/code/fairseq-constraints fairseq-interactive $modeldir \
  --bpe fastbpe \
  --bpe-codes $modeldir/bpecodes \
  --constraints \
  --constraints-both
  -s de -t en \
  --path $modeldir/model1.pt \
  --max-tokens 1000 \
  --beam 5 \
```

Adding the `--constraints-both` option causes it to batch-decode the input sentence both with and without the constraints. When run with the Fairseq WMT19 German--English model, the following results are produced (here run on a CPU, don't be alarmed by the times!)

```text
S-0     Die masch@@ in@@ elle Über@@ setzung ist schwer zu kontrollieren .
W-0     1.844   seconds
C-0     hard
C-0     influence
H-0     -1.5333266258239746     Mach@@ ine trans@@ lation is hard to influence .
D-0     -1.5333266258239746     Machine translation is hard to influence .
P-0     -0.5434 -0.1423 -0.1930 -0.1415 -0.2346 -1.8031 -0.1701 -11.7727 -0.1815 -0.1511
S-0     Die masch@@ in@@ elle Über@@ setzung ist schwer zu kontrollieren .
W-0     1.844   seconds
H-0     -0.3731671869754791     Mach@@ ine trans@@ lation is difficult to control .
D-0     -0.3731671869754791     Machine translation is difficult to control .
P-0     -0.5434 -0.1423 -0.1930 -0.1415 -0.2346 -1.1430 -0.1665 -0.8482 -0.1678 -0.1514
2020-07-31 12:17:55 | INFO | fairseq_cli.interactive | Total time: 12.803 seconds; translation time: 3.688
```

Note the new tags present in the output:

* `C-#` records active constraints (after applying preprocessing) for a sentence
* `W-#` reports the sentence-level translation time (a useful unrelated feature I hope you'll accept)

Some unit tests are written (`fairseq/test_constraints.py`) but not yet integrated. Advice here on where to place this is welcome. I also have not run this through lint; if someone can tell me the command to run, I'd appreciate it.

### Implementation notes

This is largely self-contained, implemented in a new `LexicallyConstrainedBeamSearch` class in `search.py`. It does require a few minimal hooks from `_generate()` in `sequence_generator.py`, to ensure that constraints are updated at each timestep. (Edit: most changes in that file are documentation clarifications, corrections, and updates). Unconstrained sentences that are intermingled with constrained ones will not incur any time penalty, so long as they do not occur in the same batch.

Addresses #1536.

## PR review    

## Did you have fun?


mjpost:
I'm marking this as a draft until I can straighten out all the test cases, which are giving me some trouble, in part due to the fact that I get different results running them locally.

mjpost:
Okay, code has been modified and improved such that all tests are passing.

fairseq/search.py
```diff
@@ -75,11 +100,315 @@ def step(self, step: int, lprobs, scores: Optional[Tensor]):
...
+
+        # At this point, beams_buf and indices_buf are single-dim and contain relative indices
         return scores_buf, indices_buf, beams_buf
 
 
+class LexicallyConstrainedBeamSearch(Search):
+    """Implements lexically constrained beam search by maintaining, for
+    each beam hypothesis, an object to track which of the constraints
+    have been met by that hypothesis.
+    """
```
jhcross, line 115: It might make sense to reference the paper(s) here.

fairseq/search.py
```diff
@@ -75,11 +100,315 @@ def step(self, step: int, lprobs, scores: Optional[Tensor]):
...
+        :param constraint_states: What constrained words have been generated by each item in the beam.
+        :param beams_buf: Beam numbers for each candidate. Shape: 1d list of cands.
+        :param indices_buf: Target vocab IDs for each candidate. Shape: 1d list of cands.
+        :param scores_buf: The score for each candidate. Shape: 1d list of scores.
+        """
+
+        device = lprobs.device
+
+        # STEP 2: Add all constraints for each beam item
+        for beamno, state in enumerate(constraint_states):
```
jhcross, line 291: I really enjoy pronouncing the variable name `beamno` out loud.

fairseq_cli/interactive.py
```diff
@@ -155,28 +203,45 @@ def decode_fn(x):
...
-                results.append((start_id + id, src_tokens_i, hypos))
+                constraints = list_constraints[i]
+                results.append((start_id + id, src_tokens_i, hypos,
+                                { "constraints": constraints,
+                                  "time": translate_time / len(translations) }
+                            ))
 
         # sort output to match input order
-        for id, src_tokens, hypos in sorted(results, key=lambda x: x[0]):
+        for id_, src_tokens, hypos, info in sorted(results, key=lambda x: x[0]):
```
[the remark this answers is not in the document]
mjpost: Yep!

fairseq/search.py
mjpost: Two syllables, or three?

jhcross: Two ("beam" + "no") :)

fairseq/search.py
mjpost: Will do.

fairseq/constraints.py
```diff
@@ -0,0 +1,500 @@
...
+                 node: ConstraintNode,
+                 copy_from = None):
+        self.node = node
+
+        if copy_from is None:
+            # The root node
+            self.root = node
+            # The set of states in the graph that have been completed
+            self.completed = Counter()
+            # The...
```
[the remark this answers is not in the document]
mjpost: Fixed.

fairseq/sequence_generator.py
```diff
@@ -12,6 +13,7 @@
 from fairseq.data import data_utils
 from fairseq.models import FairseqIncrementalDecoder
 from fairseq.models.fairseq_encoder import EncoderOut
+from fairseq.search import LexicallyConstrainedBeamSearch
```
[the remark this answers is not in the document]
mjpost: This is leftover from an earlier test using `isinstance` (which failed torchscript), will remove.

fairseq/sequence_generator.py
```diff
@@ -91,6 +94,10 @@ def __init__(
         # settings when the model is shared.
         self.should_set_src_lengths = hasattr(self.search, 'needs_src_lengths') and self.search.needs_src_lengths
 
+        # We only need to manage constraint states in
+        # LexicallyConstrainedBeamSearch.
+        self.constraints_active = hasattr(self.search, "constraints_active") and self.search.constraints_active
```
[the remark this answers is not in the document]
mjpost: Sure, I'll try this. I did this as a conjunct following the other example there (I assumed there was some cryptic reason for it).

fairseq/search.py
```diff
@@ -19,6 +21,9 @@ def __init__(self, tgt_dict):
         self.eos = tgt_dict.eos()
         self.vocab_size = len(tgt_dict)
         self.src_lengths = torch.tensor(-1)
+        # this test in SequenceGenerator._generate() fails if this isn't here:
+        # self.constraints_active = hasattr(self.search, "constraints_active") and self.search.constraints_active
```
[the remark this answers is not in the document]
mjpost: Will do.

fairseq/constraints.py
```diff
@@ -0,0 +1,516 @@
...
+    @staticmethod
+    def create(constraints: List[List[int]]):
+        root = ConstraintNode()
+        for sequence in constraints:
+            root.add_sequence(sequence)
+
+        return root
+
+    @staticmethod
+    def print_graph(node):
```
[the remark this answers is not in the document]
mjpost: I tried this, but it failed, because the type definition isn't yet complete. But now I'm remembering the workaround is to quote it, I think.

fairseq/constraints.py
```diff
@@ -0,0 +1,516 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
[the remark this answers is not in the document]
mjpost: Sure thing.

mjpost:
> a) instead of updating a bunch of tasks to throw if constraints are set and they are not supported there, how about we just add --constraints argument only to the translation task, which is the only one supported at the moment?

I agree the current approach is a bit cumbersome. I added all of those in order to get the test cases to pass. One counterargument is that constrained decoding has also been implemented for the Levenshtein Transformer, and there's no reason it couldn't be made to work with the multilingual decoders, too. We could add them just to these subsets, but it might be equally as messy. I'll follow your suggestion here, of course.

> b) instead of adding constraints-related code into existing classes and sprinkling if statements around, what if we had a more general solution, where we introduce hooks into existing classes (like sequence_generator.py) that call certain functions (noop by default) at various stages. then you can hook your constraint stuff through these hooks?

This would also make sense, to implement this as NOOP in the base class. The only issue I see is that, while this is more general, the nature of the hooks is pretty specific to constraints, so it might end up creating a kind of fake generality.

mjpost:
On the issue of merging `LexicallyConstrainedBeamSearch` into the base `Search` class: this makes sense, and is something I'd considered doing originally. I went this way because I thought it would be easier to merge afterwards than to split. The argument against doing it is that it does add quite a bit of complication, which can get in the way of seeing the nice simplicity of the current `Search` implementation.

mjpost:
Oh, on the hooks—yes, now I see, this makes sense. I only just added the stubs throwing `NotImplementedError` at the end, to make the test cases pass, but now that they are there, you're right it makes sense to just call them as NOOPs instead of using an `if` statement. That simplifies a lot.

alexeib:
> re: I agree the current approach is a bit cumbersome. I added all of those in order to get the test cases to pass. One counterargument is that constrained decoding has also been implemented for the Levenshtein Transformer, and there's no reason it couldn't be made to work with the multilingual decoders, too. We could add them just to these subsets, but it might be equally as messy. I'll follow your suggestion here, of course.

if its already applicable to more than one task then thats fine, we can keep the flags in options but maybe we can add the error throwing somewhere else. for example, you can add a property "supports_constraints" on the base task class that returns false, and overwrite it for tasks that do. then in some single places check that property and throw if constraints are set but task does not support them?

mjpost:
I did add this `supports_constraints` variable, defaulting to `False` in the base class, and currently only implemented in `LexicallyConstrainedBeamSearch`. The check is done in `generate()` since constraints are defined at the batch.

mjpost:
I just added an example of how to use constrained decoding under `examples/`, and then listed it in the top-level README. I think this is all set.

fairseq/options.py
```diff
@@ -609,6 +609,10 @@ def add_generation_args(parser):
                        help='sample from top K likely next words instead of all words')
     group.add_argument('--sampling-topp', default=-1.0, type=float, metavar='PS',
                        help='sample from the smallest set whose cumulative probability mass exceeds p for next words')
+    group.add_argument('--constraints', const="ordered", nargs="?", choices=["ordered", "unordered"],
+                       help='enables lexically constrained decoding')
+    group.add_argument('--constraints-both', action="store_true",
+                       help='translate sentence both constrained and unconstrained')
```
alexeib, line 615: how and why would users use this in practice? is this a common case? maybe its better to omit this + relevant code, and whoever is interested can just decode twice ?

examples/constrained_decoding/README.md
```diff
@@ -0,0 +1,129 @@
...
+    C-0     hard
+    C-0     influence
+    H-0     -1.5333266258239746     Mach@@ ine trans@@ lation is hard to influence .
+    D-0     -1.5333266258239746     Machine translation is hard to influence .
+    P-0     -0.5434 -0.1423 -0.1930 -0.1415 -0.2346 -1.8031 -0.1701 -11.7727 -0.1815 -0.1511
+
+You can also provide the `--constraints-both` flag, which will decode the sentence in the same
+batch, but unconstrained, outputting both results.
+
+    echo https://awslabs.github.io/sockeye/inference.html#lexical-constraints
```
alexeib, line 44: whats this?

README.md
```diff
@@ -40,6 +40,7 @@ We provide reference implementations of various sequence modeling papers:
   - [Neural Machine Translation with Byte-Level Subwords (Wang et al., 2020)](examples/byte_level_bpe/README.md)
   - [Unsupervised Quality Estimation for Neural Machine Translation (Fomicheva et al., 2020)](examples/unsupervised_quality_estimation/README.md)
   - [wav2vec 2.0: A Framework for Self-Supervised Learning of Speech Representations (Baevski et al., 2020)](examples/wav2vec/README.md)
+  - [Lexically constrained decoding with dynamic beam allocation (Post & Vilar, 2018)](examples/constrained_decoding/README.md)
```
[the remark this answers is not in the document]
mjpost: Fixed.

examples/constrained_decoding/README.md
mjpost: Weird typo!

fairseq/options.py
mjpost: More a debugging option. Removed.

README.md
```diff
@@ -32,6 +32,7 @@ We provide reference implementations of various sequence modeling papers:
   - [Scaling Neural Machine Translation (Ott et al., 2018)](examples/scaling_nmt/README.md)
   - [Understanding Back-Translation at Scale (Edunov et al., 2018)](examples/backtranslation/README.md)
   - [Adaptive Input Representations for Neural Language Modeling (Baevski and Auli, 2018)](examples/language_model/transformer_lm/README.md)
+  - [Lexically constrained decoding with dynamic beam allocation (Post & Vilar, 2018)](examples/constrained_decoding/README.md)
```
myleott, line 35: Can you also add a line for this in the "What's New" section below and mention constrained decoding under "Features"

examples/constrained_decoding/normalize.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python3
```
myleott, line 1: please put the fairseq license header on this and tok.py

fairseq/search.py
```diff
@@ -46,10 +49,35 @@ def step(self, step, lprobs, scores):
     def set_src_lengths(self, src_lengths):
         self.src_lengths = src_lengths
 
+    @torch.jit.export
+    def init_constraints(self, batch_constraints: Optional[Tensor], beam_size: int):
+        """Initialize constraint states for constrained decoding (if supported).
+
+        :param batch_constraints:
+        :param beam_size:
```
myleott, line 57: Please switch to the Google Python docstring convention (i.e., "Args:"), which we use this in most places in fairseq: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

fairseq/sequence_generator.py
```diff
@@ -3,6 +3,7 @@
 # This source code is licensed under the MIT license found in the
 # LICENSE file in the root directory of this source tree.
 
+import logging
```
myleott, line 6: Is logging used somewhere?

fairseq/sequence_generator.py
```diff
@@ -335,10 +346,13 @@ def _generate(
             cand_bbsz_idx = cand_beams.add(bbsz_offsets)
 
             # finalize hypotheses that end in eos
+            # Shape of eos_mask: (batch size, beam size)
```
myleott, line 349: nice, thanks for all these comments. This code is already so complicated, every bit of documentation helps :)

scripts/constraints/extract.py
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env python3
```
myleott, line 1: please add fairseq license header

PolKul:
Hi @mjpost, I'm trying to adopt your LexicallyConstrainedBeamSearch for the transformer model with the bytelevelbpe tokenizer (sub-word level tokens) but it doesn't seem to work well with it. It is copying the constraints at the beginning of the decoded sequence, then sets end-of-sequence token regardless of the input.   

It does work well with the bpe (word level tokenizer) though, producing rich outputs with the correct positions of the constraints in them. 

What do you think can be the problem with the bytelevelbpe? 

Thanks

PolKul:
My bad, I've just added eos to the constraints :) Problem solved.

mjpost:
Hmm, EOS to the constraints? Doesn't that force them to be applied at the end of the sentence? I'd be curious to understand this better, but I'm glad you have it working.

(BTW, I have found a bug with the constraint tracking. If a constraint is interrupted, instead of starting the tracking over at the beginning of that constraint, it starts over entirely. This only has an effect if you have multiple constraints. I'll have a fix in for this soon).

PolKul:
no, if I add eos to the constraints it would just output the constraints. And setting min_length doesn't help. But I have to say that I'm using it with my custom transformer model, so maybe there is something else which is affecting it. Do you have different behavior with the eos?

mjpost:
I'm not quite sure what you mean by "the eos". I assumed you meant that you were appending it to the constraints. It's hard to answer without knowing exactly what your input and command invocation are.

PolKul:
I mean that adding eos token to the end of the constrains (I have just one constrain) makes the decoder output only that constrain. Changing beam size or min_length doesn't help. I can only produce rich sentences without eos in the constrains. Let me know if you can reproduce the same...

mjpost:
If you post a minimal working example (input and command), I can take a look.

PolKul:
well, its a bit tricky to provide the working example, because, as I say, I'm using it in my custom transformer (from ParlAI library), so I have to stripe a lot of it. But basically, if you initialize it like so: 

#tokenize constraint
seed = parse(self.seed)
self.search.init_constraints([seed], beam_size)

PolKul:
hi @mjpost, I have just added my developments to the ParlAI repo. Can you please check it here https://github.com/PolKul/ParlAI

If you run tests/run_constrained_beam_search.py with your constraints list and see how it works, I would appreciate it. I see several problems:
1. Currently I cannot use more than one constraint in the list 
2. When using beam_context_block_ngram >0 it would output garbage in the generated text for the second and all consequent utterances 

Thanks for you help

PolKul:
I've started new discussion about ParlAI implementation here #https://github.com/facebookresearch/ParlAI/discussions/3582

PR #2444, opened 2020-08-07
also call model.all_reduce() in valid_step() when using LegacyDistrib…

freewym:
…utedDataParallel. Otherwise there will be a RuntimeError at [this line](https://github.com/pytorch/fairseq/blob/master/fairseq/legacy_distributed_data_parallel.py#L91) while doing validation with LegacyDDP.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq/trainer.py
+++ b/fairseq/trainer.py
@@ -643,6 +643,9 @@ class Trainer(object):
                 logging_outputs, sample_size, ignore=is_dummy_batch,
             )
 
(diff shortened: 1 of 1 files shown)


myleott:
Ah, thanks for the fix! We realized the same issue, but I think rather than all-reducing during validation we can just skip this check. It was only intended as a safety check for anybody using their own Trainer, but we don't really need the check.

PR #2455, opened 2020-08-10
RXF OS Implementation

ArmenAg:
## What does this PR do?
Implements R3F and R4F coming from Facebook Research: https://arxiv.org/abs/2008.03156

This code was used to generate all the results from the paper excluding probing results.


fairseq/models/roberta/model.py
```diff
@@ -98,6 +98,8 @@ def add_args(parser):
                             help='scalar quantization noise and scalar quantization at training time')
         parser.add_argument('--untie-weights-roberta', action='store_true',
                             help='Untie weights between embeddings and classifiers in RoBERTa')
+        parser.add_argument('--do-r4f', action='store_true',
```
myleott, line 101: Please rename this to something more descriptive, something like `--spectral-norm-classification-head`

fairseq/models/bart/model.py
```diff
@@ -59,21 +59,25 @@ def add_args(parser):
             choices=utils.get_available_activation_fns(),
             help='activation function to use for pooler layer'
         )
+        parser.add_argument('--do-r4f', action='store_true',
```
myleott, line 62: Please rename this to something more descriptive, something like `--spectral-norm-classification-head`

ArmenAg: Updated.

fairseq/models/roberta/model.py
ArmenAg: Updated.

PR #2529, opened 2020-08-26
Transformer with integrated pointer-generator network

senarvi:
This pull request implements a variant of the Transformer model that uses an attention distribution for pointing to input words. The attention distribution over the input words is interpolated with the normal output distribution over the vocabulary words, as in [See et al. (2017)](https://arxiv.org/abs/1704.04368). This allows the model to generate words that appear in the input, even if they don't appear in the vocabulary, helping especially with small vocabularies.

The mechanism for copying out-of-vocabulary words from the input has been implemented differently to See et al. In their [implementation](https://github.com/abisee/pointer-generator) they convey the word identities through the model in order to be able to produce out-of-vocabulary words. We wanted to minimize changes to the Fairseq code base and took a different approach, which I'll describe below. The entire implementation is contained in one file (plus there's one new test).

Copying out-of-vocabulary words is possible by pre-processing the input and post-processing the output. The user may add special words to the end of the vocabulary that can be used in place of `<unk>` tokens to identify different input positions (e.g. `<unk-0>`, `<unk-1>`, `<unk-2>`, ...). The number of these special words is given to the model with the `--source-position-markers` argument—the model simply maps all of these to the same word embedding as `<unk>`. With a simple post-processing the user may retrieve word at position N in the original text and use it in place of `<unk-N>`.

I didn't find a good place to document this usage of this model, so let me know if you think I should improve documentation somewhere.

This feature has not yet been discussed via a GitHub issue, but I'll open a new issue for discussion.

### DIFF
--- a/examples/pointer_generator/README.md
+++ b/examples/pointer_generator/README.md
@@ -1,0 +1,79 @@
+# Transformer with Pointer-Generator Network
+
+This page describes the transformer_pointer_generator model that incorporates a
+pointing mechanism in the Transformer model that facilitates copying of input
+words to the output. This architecture is described in [Enarvi et al. (2020)](https://www.aclweb.org/anthology/2020.nlpmc-1.4/).
+
+## Background
+
+The pointer-generator network was introduced in [See et al. (2017)](https://arxiv.org/abs/1704.04368)
+for RNN encoder-decoder attention models. A similar mechanism can be
+incorporated in a Transformer model by reusing one of the many attention
+distributions for pointing. The attention distribution over the input words is
+interpolated with the normal output distribution over the vocabulary words. This
+allows the model to generate words that appear in the input, even if they don't
+appear in the vocabulary, helping especially with small vocabularies.
+
+## Implementation
+
+The mechanism for copying out-of-vocabulary words from the input has been
+implemented differently to See et al. In their [implementation](https://github.com/abisee/pointer-generator)
+they convey the word identities through the model in order to be able to produce
+words that appear in the input sequence but not in the vocabulary. A different
+approach was taken in the Fairseq implementation to keep it self-contained in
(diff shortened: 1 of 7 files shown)


myleott:
Re: documenting, would it make sense to add a README to a new `examples/pointer_generator_networks/` directory? You could also clarify some of the usage and implementation differences there.

senarvi:
> Re: documenting, would it make sense to add a README to a new `examples/pointer_generator_networks/` directory? You could also clarify some of the usage and implementation differences there.

That's a good idea. I added documentation and some example scripts.

Also, I noticed that the pull request doesn't have the CLA Signed tag. I assume because my e-mail in the commits was in lowercase letters, while in the CLA it's capitalized. I fixed the commits (using "git commit --amend --author"), but it looks like you may have to rerun the bot.

senarvi:
@myleott could you check why this still doesn't have the CLA Signed tag?

Kyubyong:
@senarvi Can't wait for this model to be part of FairSeq. Thanks! Would you provide a concrete training script in the [README.md](https://github.com/nuance-research/fairseq/blob/transformer_pointer_generator/examples/pointer_generator/README.md) with any toy dataset? I think your <unk-0,1,..> idea is awesome. I just hope the documentation will be a little better.

senarvi:
> @senarvi Can't wait for this model to be part of FairSeq. Thanks! Would you provide a concrete training script in the [README.md](https://github.com/nuance-research/fairseq/blob/transformer_pointer_generator/examples/pointer_generator/README.md) with any toy dataset? I think your <unk-0,1,..> idea is awesome. I just hope the documentation will be a little better.

Thanks! I added this page with detailed instructions how to run this on the XSum task: https://github.com/nuance-research/fairseq/blob/transformer_pointer_generator/examples/pointer_generator/README.xsum.md

PR #2646, opened 2020-09-22
adding search.PrefixConstrainedBeamSearch

nicola-decao:
# Before submitting

## What does this PR do?
This adds a new decoding strategy `search.PrefixConstrainedBeamSearch` that limits the vocabulary of the next token generation given a prefix (that is the previously generated tokens during beam search). An end user has just to give the optional argument `prefix_allowed_tokens_fn` to `.generate` or `.sample` to activate `PrefixConstrainedBeamSearch`. `prefix_allowed_tokens_fn(batch_id, tokens)` is a callback function that given the `batch_id` and `tokens` returns the list of allowed token for the next generation step.

## Did you have fun?
YES! 🙃


nicola-decao:
The test failed on something that is not part of the pull request

fairseq/sequence_generator.py
```diff
@@ -269,6 +269,7 @@ def _generate(
 
         reorder_state: Optional[Tensor] = None
         batch_idxs: Optional[Tensor] = None
+        original_batch_idxs = sample.get("id", torch.arange(0, bsz))
```
myleott, line 272: This line seems to be causing the test failure

myleott (changes requested):
You can ignore the `test_translation_multi_simple_epoch` test failure (the `psutil` import failure has been fixed in trunk).

But the `test_ensemble_sequence_generator (tests.test_sequence_generator.TestJitSequeneceGenerator)` failures seems related (see comment below)

fairseq/sequence_generator.py
nicola-decao: I understand! But apparently the `id` field can be of different types as I got:
```
Variable 'original_batch_idxs' previously has type Optional[Tensor] but is now being assigned to a value of type Dict[str, Tensor]
```
This is weird as the filed `id` is supposed to contain an unidimensional tensor of positive integer ids.

nicola-decao: @myleott I changed a bit the type check for `original_batch_idxs`. I also had to change the signature for `search.Search.step` adding 2 optional parameters as with TorchScript it seems I cannot have conditional computation based on the type of a class. Not it should be ok. It passed the tests.

fairseq/sequence_generator.py
```diff
@@ -402,7 +395,7 @@ def _generate(
             assert num_remaining_sent >= 0
             if num_remaining_sent == 0:
                 break
-            if isinstance(self.search, search.PrefixConstrainedBeamSearch) and step >= max_len:
+            if self.search.stop_on_max_len and step >= max_len:
```
myleott, line 398: 👍

nicola-decao:
> @myleott has imported this pull request. If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/intern/diff/D24006805/).

I am not a Facebook employee so I cannot see the warnings and why this fails.

fabiopetroni:
> > @myleott has imported this pull request. If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/intern/diff/D24006805/).
> 
> I am not a Facebook employee so I cannot see the warnings and why this fails.

I'm taking care of this :)

PR #2657, opened 2020-09-25
Fix broken links to Wav2Vec2 Large checkpoints

sharvil:
Typo fixes.

PR #2679, opened 2020-10-01
Fix incorrect wav2vec model links

CatalinVoss:
Fixes links in README


myleott:
Will merge https://github.com/pytorch/fairseq/pull/2657, which has one more fix as well. Thanks for the PR!

PR #2726, opened 2020-10-12
Fix Windows CI

erip:
# Before submitting

## What does this PR do?
Adds Windows CI support so we can ensure fairseq is and remains Windows-compatible. 😄 

## PR review    

## Did you have fun?


erip:
Contrary to what CI says, the real errors arise in a difference in how garbage collection happens in Windows. Specifically, memmaps are not being released/closed appropriately due to strong references being held. The big offender is `MMapIndexedDataset`. This causes every test in `test_binaries.py` to fail (and anything that uses calls from that file).

erip (commented):
[This](https://github.com/pytorch/fairseq/pull/2726/commits/956010c4798f771fd167dd6f54f0b5d8c16acfa6) seems like a hack since "in" will not always be the source lang. I'm not positive whose responsibility it is to append the appropriate language name to the files, though. Same thing [here](https://github.com/pytorch/fairseq/pull/2726/commits/5212150f8de108bab9b1d9a82602070f271a506c#diff-79363775c289bed05f79e4cb99e1af48ebd135f9d0a632b8204249a2ef962c8dR94). Thoughts, @myleott?

erip:
The last remaining hurdle is a flaky assertion in sequence generator. 🤔

erip:
The flaky test here looks like it happens on all platforms (see [here](https://github.com/pytorch/fairseq/issues/2749) for more context). Once I address some of the hacky stuff to make tests pass here with raw datasets, hopefully this can land.

erip:
Ok, the lingering fixes are both related to how raw datasets work. Specifically it doesn't seem like the language codes get appended to the files that are read by the `RawIndexedTextDataset`. See:

- [here](https://github.com/pytorch/fairseq/pull/2726/commits/5212150f8de108bab9b1d9a82602070f271a506c#diff-79363775c289bed05f79e4cb99e1af48ebd135f9d0a632b8204249a2ef962c8dR94)
- [here](https://github.com/pytorch/fairseq/pull/2726/commits/956010c4798f771fd167dd6f54f0b5d8c16acfa6#diff-bdcc431c39fa533547a7183b9e3e3f94fb12c94e0d770cc5473d11002b585b39R198)

I think maybe it should be the responsibility of [`fairseq.data.indexed_dataset.make_dataset`](https://github.com/pytorch/fairseq/blob/master/fairseq/data/indexed_dataset.py#L57) to accept a language code (from args) and append that to the filename - this seems least invasive. Do you have any thoughts, @myleott?

erip:
Ok, I think I've got something in place for the hacky solution before.

myleott:
This looks awesome! Will import and review in a bit.

erip:
I can't peer into the black box to see where the errors are, but if there are things I can do to help it land I'm happy to.

erip:
Oh, it actually looks like the delta between this and [this PR](https://github.com/pytorch/fairseq/pull/2733) is a single linter advice so maybe this isn't such a big effort. 😄

fairseq/utils.py
```diff
@@ -452,7 +453,7 @@ def import_user_module(args):
         # a deterministic hash of the original module_path.
         with tempfile.TemporaryDirectory() as tmpdirname:
             unique_mod_name = "fairseq_user_dir_{}".format(hash(module_path) % 100000)
-            os.symlink(module_path, os.path.join(tmpdirname, unique_mod_name))
+            shutil.copytree(module_path, os.path.join(tmpdirname, unique_mod_name))
```
myleott, line 456: Ah, this is probably not what's intended -- in the case of a user passing in `--user-dir` we don't want to copy the whole thing to a temporary directory. The goal here is just to import it under a unique name, and unfortunately importlib.import_module doesn't seem to support this, so we hacked it with symlinks.

But I thought symlinks created in this way would work in Windows, is that not the case?

fairseq_cli/preprocess.py
```diff
@@ -331,7 +331,7 @@ def make_all_alignments():
         ) as f:
             for k, v in align_dict.items():
                 print("{} {}".format(src_dict[k], tgt_dict[v]), file=f)
-
+    logging.shutdown()
```
myleott, line 334: According to python docs: "This should be called at application exit and no further use of the logging system should be made after this call." However in a unittest context (or if there's library usage) we may still call into logging after this.

I assume the motivation is to close the FileHandler registered above? Can we instead save a reference to the FileHandler and do:
```
file_handler.close()
logger.removeHandler(file_handler)
```

fairseq_cli/train.py
```diff
@@ -138,6 +138,7 @@ def main(args):
         )
     train_meter.stop()
     logger.info("done training in {:.1f} seconds".format(train_meter.sum))
+    logging.shutdown()
```
myleott, line 141: do we need this one?

tests/temp_file.py
```diff
@@ -0,0 +1,29 @@
+import os
+import tempfile
+
+
+class CustomNamedTemporaryFile:
```
myleott, line 5: nice :)

tests/test_binaries.py
```diff
@@ -1375,7 +1393,9 @@ def train_masked_lm(data_dir, arch, extra_flags=None):
             "no_c10d",
             "--num-workers",
             "0",
-        ]
+            "--dataset-impl",
+            "raw",
```
myleott, line 1397: why this change? Ideally we'd test with binarized data, since that's the more common usage

myleott (changes requested):
Internal tests look good. Some questions/comments below:

fairseq/utils.py
erip: Symlinks require admin privileges in Windows, so this won't work as-is. One option might be to try to symlink and fall back to copy in the Windows case. Another is to simply make this unsupported on Windows.

fairseq/tasks/masked_lm.py
```diff
@@ -100,6 +100,11 @@ def add_args(parser):
             help="comma-separated list of dataset splits to apply shortening to, "
             'e.g., "train,valid" (default: all dataset splits)',
         )
+        parser.add_argument(
+            "--lang-code",
+            default=None,
+            help="the lang code. Only used for raw text datasets."
+        )
```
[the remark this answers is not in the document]
erip: That might be good enough, honestly. I can try to make a change and see how it works. The nature of `create_dummy_data` was a bit confusing to me, so I tried to make _something_ work. 😆

tests/test_binaries.py
erip: The reason for switching the tests to raw is because of PermissionErrors raised due to strong references held on the binary memmap'd files in `MMapIndexedDataset`. The tests consistently end before the `FairseqTask` is gc'd which means the files are still "in use" causing huge failures. My thought was that your internal tests are likely 1) mmap-based 2) Linux-only which would also test this more comprehensively.

fairseq_cli/preprocess.py
erip: Likely yes - I can update.

fairseq/tasks/masked_lm.py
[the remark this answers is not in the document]
erip: Maybe to take a step back - the reason this was introduced is because of moving to raw datasets. the issue is that unlike having a filename convention for looking up a file given a split (i.e., appending ".bin" or ".idx"), there's no such convention for raw datasets since the default extension is the bitext extensions. Without this, you'll find the tests fail with "file doesn't exist" errors for dataset `blah/blah/blah/{train,valid}`. If you _only_ specify, e.g., `/path/to/data/train.en` as your `data` arg, 1) I'm not sure dictionary loading will work - maybe it will, 2) I'm not sure how to generically infer the files corresponding to the other splits which I think is currently unhandled

fairseq/utils.py
erip: Also, this is tree copy is functionally equivalent since the directory gets cleaned up. I guess it _could_ behave differently if your `--user-dir` is, say, really really big since the copy might take some time. 😄  Are there other concerns?

tests/test_binaries.py
erip: [This commit](68d0af4a49e) adds the additional tests that are binarized but skips them on Windows. I added this to ensure no regressions are introduced on non-Windows platforms, but it would be nice to refactor these tests eventually. Probably not an immediate thing, though.

fairseq/tasks/masked_lm.py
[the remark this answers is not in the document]
erip: I actually just tested and it seems like this doesn't work - I might be doing something wrong, but I tried providing `os.path.join(data_dir, "train.in")` as the argument for positional `data` argument in the tests, but it looks like it's expected that it's a directory.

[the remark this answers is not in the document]
myleott: Sorry I'm still a bit lost :) This seems somewhat unrelated to Windows support, since Windows should still support the bin/idx files, right? I do think it'd be good to include some test coverage for raw text datasets, but would prefer to discuss that in a separate PR (and have this one just focus on Windows)

fairseq/utils.py
myleott: Ah, I see. How about a try/except and fallback to copytree if the symlink fails

fairseq/tasks/masked_lm.py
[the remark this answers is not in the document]
erip: Windows does support bin and idx, but the tests cannot prove this. It's not easy to explain, but the shortest story:

Preproces, train, validate, and generate succeed with the temp dir.

The temp dir tries to clean itself up.

Because the task gets cleaned up whenever python wants it to, the bin files haven't been closed and the test fails for that reason. This is due to very weird ownership relationships within the Datasets hierarchy.

fairseq/utils.py
erip: That seems reasonable to me!

tests/test_binaries.py
myleott: ah, I think we don't need to switch to raw for this. If you switch to `--dataset-impl=cached` everywhere in tests (including preprocessing), I think it should load the dataset into memory without the mmap.

fairseq/tasks/masked_lm.py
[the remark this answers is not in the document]
erip: FWIW, `--dataset-impl cached` is not a drop-in replacement for `--dataset-impl mmap`:

```log
tests/test_binaries.py:1271: in train_legacy_masked_language_model
    train.main(train_args)
fairseq_cli/train.py:71: in main
    task.load_dataset(valid_sub_split, combine=False, epoch=1)
fairseq/tasks/cross_lingual_lm.py:167: in load_dataset
    split=language_split, epoch=epoch
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <fairseq.tasks.cross_lingual_lm.CrossLingualLMTask object at 0x7f82a9529390>
split = 'valid.in', epoch = 1

    def _load_single_lang_dataset(self, split, epoch):
        loaded_datasets = []

        paths = utils.split_paths(self.args.data)
        assert len(paths) > 0
        data_path = paths[(epoch - 1) % len(paths)]

        for k in itertools.count():
            split_k = split + (str(k) if k > 0 else "")
            path = os.path.join(data_path, split_k)

            ds = data_utils.load_indexed_dataset(
                path, self.dictionary, self.args.dataset_impl
            )
            if ds is None:
                if k > 0:
                    break
                else:
                    raise FileNotFoundError(
>                       "Dataset not found: {} ({})".format(split, data_path)
                    )
E                   FileNotFoundError: Dataset not found: valid.in (/var/folders/v8/ls8m50ks3zz0hzq21681r8m40000gn/T/tmpcuwyf6pttest_legacy_mlm)
```

erip:
@myleott You'll see the same test failures I was observing before in 88a2249 - in the mmap dataset case, the code knows to append .idx and .bin to resolve the path for LM data, but that's not the case in other datasets.

erip:
It also doesn't seem like the datasets are being prefetched, which is causing (fatal) cache misses. 🤔

erip:
OK, @myleott - I've fixed the symlink issue and switched everything to using cached datasets. It looks like there were a few minor issues with prefetching that I've addressed, but I think this is good for a (hopefully) final review whenever. 😺

erip:
Any chance of importing this to make sure I haven't broken anything, @myleott? If a review has to wait a bit, that's OK.

erip:
I've cleaned this up so it's hopefully easier to review. Let me know if you need anything else from me, @myleott

erip:
Any thoughts here, @myleott? The only "core" fairseq changes (i.e., not changes to tests) are: 

- closing the file handler
- fixing dtypes to `int64`
- prefetching fixes

erip:
Any chance you could review/merge this, @myleott or @alexeib? It's been imported a handful of times without review so I have to keep fixing conflicts when upstream diverges.

myleott:
Sorry for the delay and thanks for rebasing.

A quick status update on this. We'd like to avoid duplicated versions of each test (one with mmap and another with cached), so we'll probably merge a simplified version of this diff if we can get mmap working on Windows. Having looked into it a bit, the problem seems to be that Windows refuses to delete the temporary directories in the tests while the mmap is open, and technically the tempfile.TemporaryDirectory context manager closes before the mmap is garbage collected.

I think tests should work if we do something like this instead:
```python
class TestLanguageModeling(unittest.TestCase):
    def setUp(self):
        logging.disable(logging.CRITICAL)
        self._tmp_dir = tempfile.TemporaryDirectory()
        self.data_dir = self._tmp_dir.name

    def tearDown(self):
        logging.disable(logging.NOTSET)
        self._tmp_dir.cleanup()

    # everything else as before, but use `self.data_dir` instead of the context manager
```

I started along this path and things seem to be working, just need to find some time to get it over the finish line. I'll try to merge it later in the week.

Longer term, to set expectations, I do worry that Windows support will continue to break, since I don't think any of our production or research teams use Windows 😢 But in any case having CI coverage will be helpful.

erip:
> I think tests should work if we do something like this instead ...

Clever. :-) I'm happy to rework that if you'd like.

> Longer term, to set expectations, I do worry that Windows support will continue to break

Undoubtedly yes. My hope is to begin adding tests to increase coverage to the point that we have some confidence in cross-platform parity. Adding CI is a first-step and I hope that it operates as due diligence for Windows compatibility much like every other aspect of CI; i.e., if some tests fail, the PR cannot land.

myleott:
> Clever. :-) I'm happy to rework that if you'd like.

Ah thanks, that would be very helpful. So basically revert to the original mmap-based tests everywhere, but using this new tearDown pattern 😄

erip:
I don't know why pycharm dorked up the formatting, but if you can share your black config I can reformat as a part of this. ☹️

erip:
The failures are because of temp directories that are created in multilingual transformer tests. I've mis-ported these because it's not a static data directory like other tests, but a data directory per lang pair. I need to give this some more thought...

myleott:
> The failures are because of temp directories that are created in multilingual transformer tests. I've mis-ported these because it's not a static data directory like other tests, but a data directory per lang pair. I need to give this some more thought...

Ah, you can move this to a new test class

erip:
It still doesn't work on Windows ☹️ I'll give this another look later, but mmap is causing a lot of issues.

erip:
The current strategy is quite hacky, but basically it defers temp directory cleanup until gc by which point it seems like the task has been gc'd, too.

erip:
Finally. 😅  OK, @myleott - this change now has a much smaller surface area.

erip:
OK, this has been rebased and seems to be working again.

erip:
This failure is somewhat mysterious to me - @alexeib or @myleott, do you have any idea why apex would fail to be found in my tests but not yours?

erip:
Oh, I think it's because my test runner in this PR (`pytest`) discover the tests in the submodule and yours (`python setup.py test`) does not. By discovering it, it tries to import something with a strict dependency on apex. This seems like a bug in master currently.

erip:
Is this waiting on anything?

erip:
Now that the binary tests have been reworked to support mmap, are there any other blockers here @myleott?

erip:
Sorry to keep harassing you all, but it takes a fair bit of mental effort to keep this branch up to date and to fix conflicts and other issues resulting with upstream changes. Are there plans to eventually merge this or should I abandon it, @myleott?

erip:
Since this has lost interest from maintainers, I'm going to assume there's no hope of this landing.

PR #1781, opened 2020-03-05
Weights and Biases support

Guitaricet:
add WandBProgressBarWrapper
add --wandb-project argument
update progress_bar to support wandb_project

# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [x] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/master/CONTRIBUTING.md)?
- [ ] Did you make sure to update the docs?   
- [ ] Did you write any new necessary tests?  

## What does this PR do?

Adds Weights and Biases support for logging. 

## PR review    
Anyone in the community is free to review the PR once the tests have passed.     
If we didn't discuss your PR in Github issues there's a high chance it will not be merged.

## Did you have fun?
Make sure you had fun coding 🙃

PR #2753, opened 2020-10-19
Reproduce #1781. Add Weights and Biases support

### LINKED ISSUE #1790: Weights and Biases support
## 🚀 Feature Request
Add [WandB](https://www.wandb.com/) support for logging.

### Motivation

- WandB is cloud-based - you can access your logs without a direct ssh connection, just using the website
- WandB interface is better suited for many experiments, it also can log model weights, configs, and source code
- WandB is prettier

### Pitch

Tensorboard is slow a buggy - sometimes I wait about a minute for it to load all the logs, sometimes it just doesn't show them.
There is no reason not to allow fairseq users to use wandb.
It is also very simple to add a new progress bar wrapper (and I already did that).

### Alternatives

Logger extension interface (plug-in) which can allow users to use any logging method.

### Additional context

I already coded this feature for myself. Currently, I am using it for my experiments and it looks pretty good. You can see PR draft  #1781 for more information.
I'll be happy if you can help me to understand why Windows build is failing, I did not write any platform-specific code and I have no windows computer to test it.

CrafterKolyan:
# Before submitting

- [x] Was this discussed/approved via a Github issue? issue #1790

## What does this PR do?
Fixes #1790.

## PR review    

## Did you have fun?


fairseq/sequence_generator.py
```diff
@@ -421,7 +421,7 @@ def _generate(
                 break
             if self.search.stop_on_max_len and step >= max_len:
                 break
-            assert step < max_len
+            assert step <= max_len
```
myleott, line 424: is this change related?

myleott:
Thanks for contributing this!

fairseq/sequence_generator.py
CrafterKolyan: No, it is not, but the tests are constantly failing in all pull requests due to this line and I actually understand the logic behind `step < max_len`. Why it can't be equal to `max_len` at that point.

Guitaricet:
Thank you for adding wandb support! 🎉 
It would be great if we can find a way to log all the hyperparameters. This is super important for the reproducibility and, in my opinion, is an essential part of wandb.

Initially, this was implemented via adding optional parameter `args` to `progress_bar` function and `WandBProgressBarWrapper` constructor, but if you don't think this is a good idea, we can figure out some other way. Maybe via updating `wandb.config` before the train loop starts.

CrafterKolyan:
@Guitaricet it saves command. In my opinion that's enough 
image

CrafterKolyan:
@Guitaricet it saves command. In my opinion that's enough
image

Guitaricet:
Command saving is really good, but the upsides of logging the actual hparams are:

* You can search for particular hparams in wandb and plot with legend that shows hparams. This is one of the main reasons why I, personally, use wandb.
* You can compare the runs in wandb interface.
* All hparams are logged, even if you didn't specify them explicitly. This may help a lot if an architecture object changes, plus you don't need to look into the code to figure out the hparams if you need them.

myleott:
Merged in 1a709b2a401ac8bd6d805c8a6a5f4d7f03b923ff

PR #2815, opened 2020-10-29
Simplify --user-dir and require user-dir module name to be globally unique

myleott:
This PR reverts recent changes that attempted to make `--user-dir` work with non-unique module names. But that new approach introduced other issues (e.g., poor compatibility with multiprocessing and Windows), so let's revert to the previous simpler implementation.


huangxt39:
wow, the update was very helpful! It solved the problem I just encountered. In the previous version, if I use my own task file and train on multiple GPUs it will raise ModuleNotFoundError: No module named 'fairseq_user_dir_11950'.

PR #2835, opened 2020-11-03
Noise mixer

mychiux413:
## What does this PR do?
Hi,
This PR is an augmentation option to mix noise audio in process `RawAudioDataset.post_process()`,
sometimes we want to apply our ASR model just in some specific sound environments,
to make the model more adaptive to the target environment,
mixing the target environment sound into datasets while training should be better than `SpecAugment` to decrease `WER`.
 

Usage:
you can add the args below when execute `train.py` or `example/speech_recognition/infer.py`:
`--train-noise-dir`: str, A directory, put the soundfile readable files into it, the progress will automatically mix them into train datasets.
`--valid-noise-dir`: str, A directory, put the soundfile readable files into it, the progress will automatically mix them into valid datasets.
`--test-noise-dir`: str, A directory, put the soundfile readable files into it, the progress will automatically mix them into test datasets.
`--noise-min-snr-db`: float, default=3, The minimum signal-to-noise ratio between audio and noise.
`--noise-max-snr-db`: float, default=40, The maximum signal-to-noise ratio between audio and noise.

P.S. To test if the noise really mixed into datasets properly or not, we should hack into `RawAudioDataset.post_process` to dump the mixed feats out by `sf.write`, and listen it by our EARS.

## PR review    

## Did you have fun?
YES!

### DIFF
--- a/fairseq/data/audio/raw_audio_dataset.py
+++ b/fairseq/data/audio/raw_audio_dataset.py
@@ -11,17 +11,11 @@ import sys
 import numpy as np
 import torch
 import torch.nn.functional as F
-try:
-    import soundfile as sf
-except ImportError as err:
-    print(err)
-    print('try: pip install soundfile')
 
 from .. import FairseqDataset
 
 
 logger = logging.getLogger(__name__)
-DBFS_COEF = 10.0 / torch.log10(torch.scalar_tensor(10.0))
 
 
 class RawAudioDataset(FairseqDataset):
@@ -34,7 +28,6 @@ class RawAudioDataset(FairseqDataset):
         min_length=0,
         pad=False,
         normalize=False,
-        wav_augment=None,
     ):
         super().__init__()
 
@@ -48,13 +41,6 @@ class RawAudioDataset(FairseqDataset):
         self.pad = pad
         self.shuffle = shuffle
         self.normalize = normalize
-
-        self.chain = None
-        self.src_info = {"rate": sample_rate}
-        self.target_info = {"rate": sample_rate}
-        if wav_augment:
-            from fairseq.data.audio.wav_augment import chain_fatory
-            self.chain = chain_fatory(wav_augment)
 
     def __getitem__(self, index):
         raise NotImplementedError()
@@ -70,11 +56,6 @@ class RawAudioDataset(FairseqDataset):
             raise Exception(f"sample rate: {curr_sample_rate}, need {self.sample_rate}")
 
         assert feats.dim() == 1, feats.dim()
-
-        if self.chain:
-            feats = self.chain.apply(feats.unsqueeze(
-                0), self.src_info, self.target_info)
-            feats = feats[0, :]
 
         if self.normalize:
             with torch.no_grad():
@@ -160,7 +141,6 @@ class FileAudioDataset(RawAudioDataset):
         min_length=0,
         pad=False,
         normalize=False,
-        wav_augment=None,
     ):
         super().__init__(
             sample_rate=sample_rate,
@@ -170,7 +150,6 @@ class FileAudioDataset(RawAudioDataset):
             min_length=min_length,
             pad=pad,
             normalize=normalize,
-            wav_augment=wav_augment,
         )
 
         self.fnames = []
@@ -190,6 +169,8 @@ class FileAudioDataset(RawAudioDataset):
         logger.info(f"loaded {len(self.fnames)}, skipped {skipped} samples")
 
     def __getitem__(self, index):
+        import soundfile as sf
+
         fname = os.path.join(self.root_dir, self.fnames[index])
         wav, curr_sample_rate = sf.read(fname)
         feats = torch.from_numpy(wav).float()
--- a/fairseq/data/audio/raw_audio_dataset.py
+++ b/fairseq/data/audio/raw_audio_dataset.py
@@ -104,6 +104,9 @@ class RawAudioDataset(FairseqDataset):
 
         if self.noise_files_len:            
             noise: torch.Tensor = self.sample_noise()
+
+            # peak protection
+            max_gain_ratio = 1.0 / torch.max(noise)
             
             noise = noise.repeat(np.math.ceil(len(feats) / len(noise)))
             delta_len = len(noise) - len(feats)
@@ -120,7 +123,7 @@ class RawAudioDataset(FairseqDataset):
             gain_db = current_snr_db - target_snr_db
             gain_ratio = torch.pow(10, gain_db / 20)
 
-            feats += seg_noise * gain_ratio
+            feats += seg_noise * min(gain_ratio, max_gain_ratio)
 
         if self.normalize:
             with torch.no_grad():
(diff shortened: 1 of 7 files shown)


alexeib:
we have some internal impl based on https://github.com/facebookresearch/WavAugment, which i believe can also do additive noise. i have a dataset that can be used to wrap around audio dataset to add the noise, like this:

```
# Copyright (c) Facebook, Inc. and its affiliates.
#
# This source code is licensed under the MIT license found in the

import numpy as np
import torch

from .. import BaseWrapperDataset


class ChainRunner(object):
    """
    Takes an instance of augment.EffectChain and applies it on pytorch tensors.
    """

    def __init__(self, chain):
        self.chain = chain

    def __call__(self, x):
        """
        x: torch.Tensor, (channels, length). Must be placed on CPU.
        """
        x = x.view(1, -1)
        src_info = {'channels': x.size(0),  # number of channels
                    'length': x.size(1),  # length of the sequence
                    'precision': 32,  # precision (16, 32 bits)
                    'rate': 16000.0,  # sampling rate
                    'bits_per_sample': 32}  # size of the sample

        target_info = {'channels': 1,
                       'length': x.size(1),
                       'precision': 32,
                       'rate': 16000.0,
                       'bits_per_sample': 32}

        y = self.chain.apply(x, src_info=src_info, target_info=target_info)

        # sox might misbehave sometimes by giving nan/inf if sequences are too short (or silent)
        # and the effect chain includes eg `pitch`
        if torch.isnan(y).any() or torch.isinf(y).any():
            return x
        return y.squeeze(0)


class AudioAugmentDataset(BaseWrapperDataset):
    def __init__(self, dataset):
        super().__init__(dataset)

        import augment
        effect_chain = augment.EffectChain() \
            .pitch("-q", self.random_pitch_shift) \
            .rate("-q", 16_000) \
            .reverb(50, 50, self.random_room_size).channels() \
            .time_dropout(max_seconds=50. / 1_000)
        # .additive_noise(noise_generator, snr=0, denormalize_noise=False) \

        self.runner = ChainRunner(effect_chain)

    def random_pitch_shift(self):
        return np.random.randint(-300, 300)

    def random_room_size(self):
        return np.random.randint(0, 100)

    def __getitem__(self, index):
        item = self.dataset[index]
        item["original_source"] = item["source"]
        item["source"] = self.runner(item["source"])

        return item
```

do you think it makes sense to rework your pr to use this kind of approach instead?

mychiux413:
That would be great to introduce WavAugment into fairseq framework, and it has even more augment functions.

fairseq/data/audio/raw_audio_dataset.py
```diff
@@ -42,6 +49,13 @@ def __init__(
         self.shuffle = shuffle
         self.normalize = normalize
 
+        self.chain = None
+        self.src_info = {"rate": sample_rate}
+        self.target_info = {"rate": sample_rate}
+        if wav_augment:
+            from fairseq.data.audio.wav_augment import chain_fatory
```
alexeib, line 56: "fatory"?

fairseq/data/audio/wav_augment.py
```diff
@@ -0,0 +1,151 @@
...
+        else:
+            raise "{} parsing error".format(value)
+    elif re.findall(r'^[0-9\-]+$', value):
+        value = int(value)
+    elif re.findall(r'^[\.0-9\-]+$', value):
+        value = float(value)
+    return value
+
+
+def chain_fatory(commands) -> CustomEffectChain:
```
alexeib, line 123: fatory?

alexeib (changes requested):
this is looking good! could you please move the augment functionality into a wrapper dataset (like in my previous comment) and then choose to wrap or not wrap a dataset in audio_pretraining task based on a flag

something like this:

```
        manifest = os.path.join(self.args.data, "{}.tsv".format(split))
        self.datasets[split] = FileAudioDataset(
            manifest,
            sample_rate=self.args.sample_rate,
            max_sample_size=self.args.max_sample_size,
            min_sample_size=self.args.max_sample_size,
            min_length=self.args.min_sample_size,
            pad=self.args.labels is not None or self.args.enable_padding,
            normalize=self.args.normalize,
            rescale=getattr(self.args, 'rescale', False),
            speakers=self.speakers,
        )

        if getattr(self.args, 'augment', False) and split not in self.args.valid_subset:
            self.datasets[split] = AudioAugmentDataset(self.datasets[split])
```

fairseq/data/audio/raw_audio_dataset.py
```diff
@@ -11,11 +11,17 @@
 import numpy as np
 import torch
 import torch.nn.functional as F
+try:
+    import soundfile as sf
+except ImportError as err:
+    print(err)
+    print('try: pip install soundfile')
```
alexeib, line 18: this is going to get printed out for other people who dont use speech related functionality, since we read all datasets in data/__init_\_

fairseq/data/audio/wav_augment.py
```diff
@@ -0,0 +1,151 @@
...
+import torch
+import numpy as np
+import re
+from functools import partial
+
+try:
+    from augment import EffectChain
+except ImportError as err:
+    err_msg = str(err)
+    err_msg += "\nTo apply the wav_augment, you should install the `facebookresearch/WavAugment` first.\n" + \
```
alexeib, line 15: also this - maybe import in init()?

fairseq/data/audio/wav_augment.py
mychiux413: fixed

fairseq/data/audio/raw_audio_dataset.py
mychiux413: fixed

fairseq/data/audio/wav_augment.py
mychiux413: fixed

fairseq/data/audio/raw_audio_dataset.py
mychiux413: fixed

mychiux413:
> this is looking good! could you please move the augment functionality into a wrapper dataset (like in my previous comment) and then choose to wrap or not wrap a dataset in audio_pretraining task based on a flag
> 
> something like this:
> 
> ```
>         manifest = os.path.join(self.args.data, "{}.tsv".format(split))
>         self.datasets[split] = FileAudioDataset(
>             manifest,
>             sample_rate=self.args.sample_rate,
>             max_sample_size=self.args.max_sample_size,
>             min_sample_size=self.args.max_sample_size,
>             min_length=self.args.min_sample_size,
>             pad=self.args.labels is not None or self.args.enable_padding,
>             normalize=self.args.normalize,
>         )
> 
>         if getattr(self.args, 'augment', False) and split not in self.args.valid_subset:
>             self.datasets[split] = AudioAugmentDataset(self.datasets[split])
> ```

I implemented the AudioAugmentDataset, but wrapping FileAudioDataset can make the `normalize` processing before `WavAugment`, I'm not sure whether it will affect the model while training with normalize.

alexeib:
i believe wavaugment expects audio data to be normalized, and also rescaled to [-1,1] (which --normalize currently doesnt do). when i was playing around with this i was doing something like this:

```
        if self.normalize or self.rescale:
            with torch.no_grad():
                feats = F.layer_norm(feats, feats.shape)
                if self.rescale:
                    feats = feats / feats.abs().max()
```

PR #2881, opened 2020-11-10
deprecation warning fixes

stas00:
## What does this PR do?

Fixes:

- 2x `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working`

- 1x `/fairseq/optim/adam.py:98: DeprecationWarning: invalid escape sequence \:`

This is with py38.


stas00:
CLA has been signed

PR #2903, opened 2020-11-17
Metrics

janchorowski:


myleott:
Sorry, not sure I understand what's going on in this PR. It looks like you're adding handwriting recognition? That seems pretty cool, but please first open an Issue to discuss large new features. In this case, we'd most likely want to keep the implementation self-contained in the examples/ directory (similar to examples/speech_recognition) and ideally the implementation would reproduce results from a paper. Closing for now, but we'd be excited about this if you're open to refactoring things a bit! 😄

janchorowski:
Hi, 

yes we are working on handwriting recognition with wav2vec2.0, but this was meant to be against our own fork, until things work.

Sorry for spamming you, we'll send the PR with a proper intro and issue when things do work.

PR #2914, opened 2020-11-18
v0.10.1

myleott:


erip:
@myleott would you entertain a PR to also package a conda artifact on new tags, too? not sure what the exact process is for publishing to the pytorch conda channel, but I can do some of the legwork there.

myleott:
@erip yeah that sounds good.

I'm actually reworking this a bit, since auto-uploading the wheels seems a bit risky right now. Instead I'm using the [`upload-artifact` action](https://github.com/fairseq/fairseq/commit/8d66c7c0b2afa612a974714248ab2daa156bce2e#diff-52d0610b43ce35e44510ae2d15d70668334378b01c1bdc774159cd3a33b71728) to produce the artifacts and then they can easily be downloaded and uploaded manually (example: https://github.com/fairseq/fairseq/runs/1419699104).

Assuming conda works similarly, it'd be great to have some automated build process to build conda packages in a similar way.

PR #3149, opened 2021-01-21
Use torch pipe if available in fairseq.

pritamdamania87:
Summary: fairscale.nn.Pipe has been ported to PyTorch:
https://github.com/pytorch/pytorch/blob/master/torch/distributed/pipeline/sync/pipe.py#L138.
As a result, modifying the pipeline transformer to use PyTorch pipe if available. This change depends on https://github.com/pytorch/pytorch/pull/50860.

Test Plan:
```
python train.py ru_en_bin/ --arch transformer_iwslt_de_en_pipeline_parallel --share-decoder-input-output-embed --optimizer adam --adam-betas '(0.9, 0.98)' --clip-norm 0.0 --lr 5e-4 --lr-scheduler inverse_sqrt --warmup-updates 4000 --dropout 0.3 --weight-decay 0.0001 --criterion label_smoothed_cross_entropy --label-smoothing 0.1 --max-tokens 4096 --eval-bleu --eval-bleu-args '{"beam": 5, "max_len_a": 1.2, "max_len_b": 10}' --eval-bleu-detok moses --eval-bleu-remove-bpe --eval-bleu-print-samples --best-checkpoint-metric bleu --maximize-best-checkpoint-metric --pipeline-model-parallel --pipeline-balance '[1,3,5,3,3,1]' --pipeline-devices '[0,1,0,2,3,0]' --pipeline-chunks 16 --distributed-world-size 1 --distributed-no-spawn --disable-validation --max-epoch 1
```

Output with torch pipe:
```
2021-01-20 16:13:35 | INFO | train | epoch 001 | loss 12.676 | nll_loss 12.331 | ppl 5151.97 | wps 5108 | ups 1.66 | wpb 3081.6 | bsz 131.6 | num_updates 380 | lr 4.75e-05 | gnorm 2.08 | train_wall 229 | wall 233
2021-01-20 16:13:36 | INFO | fairseq_cli.train | done training in 233.1 seconds
```

Output with fairscale pipe:
```
2021-01-20 14:13:59 | INFO | train | epoch 001 | loss 12.677 | nll_loss 12.331 | ppl 5152.07 | wps 5198.9 | ups 1.69 | wpb 3081.6 | bsz 131.6 | num_updates 380 | lr 4.75e-05 | gnorm 2.08 | train_wall 224 | wall 228
2021-01-20 14:13:59 | INFO | fairseq_cli.train | done training in 228.0 seconds
```

### DIFF
--- a/fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
+++ b/fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
@@ -39,15 +39,38 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_MAX_SOURCE_POSITIONS = 1024
 DEFAULT_MAX_TARGET_POSITIONS = 1024
+TORCH_PIPE = False
 
 
 @register_model("pipeline_parallel_transformer")
 class PipelineParallelTransformerModel(BaseFairseqModel):
     def __init__(self, encoder, decoder, balance, devices, chunks, checkpoint):
         try:
-            from fairscale.nn import Pipe
+            from torch.distributed.pipeline.sync import Pipe
+            from torch.distributed.pipeline.sync.utils import partition_model
+            global TORCH_PIPE
+            TORCH_PIPE = True
+
+            # Initialize single process RPC agent since TORCH_PIPE requires
+            # RRef.
+            from torch.distributed import rpc
+            import tempfile
+            tmpfile = tempfile.NamedTemporaryFile()
+            rpc.init_rpc(
+                name="worker",
+                rank=0,
+                world_size=1,
+                rpc_backend_options=rpc.TensorPipeRpcBackendOptions(
+                    init_method="file://{}".format(tmpfile.name),
+                )
+            )
+            logger.info('Using torch pipe')
         except ImportError:
-            raise ImportError("Please install fairscale with: pip install fairscale")
+            try:
+                from fairscale.nn import Pipe
+                logger.info('Using fairscale pipe')
+            except ImportError:
+                raise ImportError("Please install fairscale with: pip install fairscale")
         super().__init__()
         assert isinstance(encoder, FairseqEncoder)
         assert isinstance(decoder, FairseqDecoder)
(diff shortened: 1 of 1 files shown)


fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
```diff
@@ -87,7 +117,10 @@ def forward(self, src_tokens, src_lengths, prev_output_tokens):
         if self.training:
             input_lst = [src_tokens, src_lengths, prev_output_tokens]
             input = tuple(i.to(self.devices[0], non_blocking=True) for i in input_lst)
-            return self.model(input)
+            if TORCH_PIPE:
+                return self.model(input).local_value()
```
shruti-bh, line 121: What does `.local_value()` do?

fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
```diff
@@ -39,15 +39,38 @@
...
+    # RRef.
+    tmpfile = tempfile.NamedTemporaryFile()
+    rpc.init_rpc(
+        name="worker",
+        rank=0,
+        world_size=1,
+        rpc_backend_options=rpc.TensorPipeRpcBackendOptions(
+            init_method="file://{}".format(tmpfile.name),
+        )
+    )
```
shruti-bh, line 59: Can you add some more details on why Torch Pipe needs this rpc.init_rpc() call?

pritamdamania87: Updated the comment on line 49 for more information. Basically, the pipe returns an RRef which depends on RPC being initialized.

fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
pritamdamania87: PyTorch Pipe returns an [RRef](https://pytorch.org/docs/stable/rpc.html#rref) pointing to the output of the pipeline. This was introduced in the API to support cross host pipelines in the future. For now, the pipe implementation is only a single host so we can retrieve the value using `local_value` method of the RRef since it will always be local. However, in the future the output of the pipeline could be on a different host and we probably need methods like `to_here` on the RRef to retrieve the remote value.

fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
```diff
@@ -561,10 +605,10 @@ def __init__(
...
         super().__init__(dictionary)
         self.register_buffer("version", torch.Tensor([3]))
-        try:
-            from fairscale.nn import Pipe
-        except ImportError:
-            raise ImportError("Please install fairscale with: pip install fairscale")
+        if TORCH_PIPE:
+            logger.info('Using torch pipe')
+        else:
+            logger.info('Using fairscale pipe')
```
msbaines, line 611: Earlier, the ImportError was thrown on use. With this change, just importing this file is sufficient to cause an import error.

msbaines (commented):
LGTM other than one concern. Will leave final approval to Shruti.

fairseq/model_parallel/models/pipeline_parallel_transformer/model.py
shruti-bh: @pritamdamania87 - could you add the import statements within the `__init__()` ?

shruti-bh:
Looks great to me apart from @msbaines remaining comment

pritamdamania87:
@shruti-bh Addressed @msbaines's comment, could you take another look? Thanks!

PR #3198, opened 2021-02-02
Update dynamic_convolution.py

JonnesLin:
There are a lot of default values and I just need to set the input_size in theory. However, the default value of the argument padding_l is None and it will lead an error. So i think you can take the default value of the argument to 0 for preventing errors.

# Before submitting

## What does this PR do?
Fixes # (issue).

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq/modules/dynamic_convolution.py
+++ b/fairseq/modules/dynamic_convolution.py
@@ -96,7 +96,7 @@ class DynamicConv1dTBC(nn.Module):
         self,
         input_size,
         kernel_size=1,
-        padding_l=None,
+        padding_l=0,
         num_heads=1,
         weight_dropout=0.0,
         weight_softmax=False,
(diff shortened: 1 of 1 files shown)


fairseq/modules/dynamic_convolution.py
```diff
@@ -16,7 +16,7 @@
 def DynamicConv(
     input_size,
     kernel_size=1,
-    padding_l=None,
+    padding_l=1,
```
myleott, line 19: should this be 0?

myleott (changes requested):

fairseq/modules/dynamic_convolution.py
JonnesLin: It can be any value greater than or equal to 0. However, it can't be None.

Issue #3194, opened 2021-02-01
pip install on Windows gives permission errors accessing a temp folder through an Admin console


munael:
## 🐛 Bug

`pip install [-U [--force-reinstall]] fairseq` currently gives `PermissionError: [WinError 5] Access is denied: 'fairseq\\examples'`. I have `0.9` installed, and trying to install `0.10`.

### To Reproduce

Steps to reproduce the behavior (**always include the command you ran**):

1. Open an admin console.
2. Run cmd `pip install -U fairseq`
3. See error. `fairseq/examples` is under:
```
cwd: $HOME\AppData\Local\Temp\pip-install-18nls8px\fairseq
```

Stack trace:
<detail>
```
  ERROR: Command errored out with exit status 1:
   command: '$HOME\AppData\Local\Continuum\anaconda3\envs\dev\python.exe' '$HOME\AppData\Local\Continuum\anaconda3\envs\dev\lib\site-packages\pip\_vendor\pep517\_in_process.py' get_requires_for_build_wheel '$HOME\AppData\Local\Temp\tmpprijkq_f'
       cwd: $HOME\AppData\Local\Temp\pip-install-18nls8px\fairseq
  Complete output (31 lines):
  Traceback (most recent call last):
    File "setup.py", line 214, in <module>
      do_setup(package_data)
    File "setup.py", line 191, in do_setup
      zip_safe=False,
    File "$HOME\AppData\Local\Temp\pip-build-env-yuwvot27\overlay\Lib\site-packages\setuptools\__init__.py", line 152, in setup
      _install_setup_requires(attrs)
    File "$HOME\AppData\Local\Temp\pip-build-env-yuwvot27\overlay\Lib\site-packages\setuptools\__init__.py", line 147, in _install_setup_requires
      dist.fetch_build_eggs(dist.setup_requires)
    File "$HOME\AppData\Local\Temp\pip-build-env-yuwvot27\overlay\Lib\site-packages\setuptools\build_meta.py", line 60, in fetch_build_eggs
      raise SetupRequirementsError(specifier_list)
  setuptools.build_meta.SetupRequirementsError: ['cython', 'numpy', 'setuptools>=18.0']

  During handling of the above exception, another exception occurred:

  Traceback (most recent call last):
    File "$HOME\AppData\Local\Continuum\anaconda3\envs\dev\lib\site-packages\pip\_vendor\pep517\_in_process.py", line 280, in <module>
      main()
[...]

PR #3380, opened 2021-03-20
A quick fix to make pip installation on Windows 10 work.

sierkov:
This is a proposed quick fix for #3194 (https://github.com/pytorch/fairseq/issues/3194)
On Windows 10, creating a symlink requires super user priviledges, so I changed the code to use a copytree instead.
This seems OK since the data volume to be copied is small. LMK

No tests were added since this an installation script and my goal is to show that a simple solution to the issue exists.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import os
+import shutil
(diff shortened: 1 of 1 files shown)


nihirv:
Still relevant

PR #3427, opened 2021-04-01
hub_utils: fix logging of src tokens

raphaelmerx:
tokens logging was using the tgt_dict on src tokens

## Description
In `GeneratorHubInterface.generate`, we attempt to log the stringified source tokens in `logger.info("S\t{}".format(self.string(source_tokens)))`. However `self.string` uses the `tgt_dict` for the conversion, so we're applying the `tgt_dict` to source tokens! This PR fixes that bug by using the `src_dict` on the logged source tokens.


tdtvbds:
it great thanks so much

raphaelmerx:
bump

PR #3557, opened 2021-05-17
call cpu conversion to pack_padded_sequence input_lengths for berard …

firaschaabani:
###  this PR fixes the Error "nn.utils.rnn.pack_padded_sequence: RuntimeError: 'lengths' argument should be a 1D CPU int64 tensor, but got 1D cuda:0 Long tensor" when using berard architecture
the **lengths** parameter of pack_padded_sequence()  _"must be in cpu if provided as a tensor" even when using GPU (CUDA) _  as mentioned in the documentation [documentation](https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pack_padded_sequence.html) and in issue pytorch/pytorch#43227.


firaschaabani:
This issue was solved by @myleott in  pytorch/pytorch#43227 i think that this PR can be aproved by you.

PR #3560, opened 2021-05-20
enable pin_memory for DataLoaders

msbaines:
To avoid the creation of a cuda:0 context, I needed to make sure that the `BackgroundConsumer` thread had its cuda device context set to the correct GPU.

### DIFF
--- a/fairseq/data/iterators.py
+++ b/fairseq/data/iterators.py
@@ -214,6 +214,7 @@ class StreamingEpochBatchIterator(EpochBatchIterating):
             num_workers=self.num_workers,
             timeout=self.timeout,
             worker_init_fn=worker_init_fn,
+            pin_memory=True,
         )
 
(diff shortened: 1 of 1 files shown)


sshleifer (approved):
For my learning, what is cuda:0 context?

msbaines:
> For my learning, what is cuda:0 context?

In torch, CUDA GPU0 is referred to as device="cuda:0" so I really meant a CUDA context for GPU 0. pin_memory will create a context for the current CUDA device which defaults to GPU 0. To avoid this, you can change the default device via torch.set_device(). The problem is the the default device is thread-specific, so new thread fallback to GPU 0 unless you call torch.set_device() on the thread.

Some more context:

https://github.com/pytorch/pytorch/issues/58626
https://github.com/pytorch/pytorch/issues/32424

PR #3592, opened 2021-06-04
fix logging when running single-process

msbaines:
In file_io.py, there is a logging message that happens in the global
scope. This logging message can be invoked before calling
logging.basicConfig() in fairseq_cli/train.py resulting in that
call becoming a no-op. This was causing the loglevel to remain at
WARNING.

Fix is to call logging.basicConfig() before import-ing any fairseq
libraries that may do logging in global scope.

Verified that I logging.info messages are now visible after applying
this PR.

# Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/master/CONTRIBUTING.md)?

## What does this PR do?
Fixes # (issue).

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq_cli/train.py
+++ b/fairseq_cli/train.py
@@ -14,6 +14,15 @@ import os
 import sys
 from typing import Dict, Optional, Any, List, Tuple, Callable
 
+# We need to setup root logger before importing any fairseq libraries.
+logging.basicConfig(
+    format="%(asctime)s | %(levelname)s | %(name)s | %(message)s",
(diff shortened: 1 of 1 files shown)


msbaines:
From pytorch docs:


> The functions debug(), info(), warning(), error() and critical() will call basicConfig() automatically if no handlers are defined for the root logger.
> 
> This function does nothing if the root logger already has handlers configured, unless the keyword argument force is set to True.


https://docs.python.org/3/library/logging.html#logging.basicConfig

sujitoc (approved):

PR #3601, opened 2021-06-08
[Hubert] Enabled storing of dictionaries

neeyanthkvk:
## What does this PR do?
For HubertPretrainingTask, added dictionaries to the task state to enable the serialization of the dictionaries (thus removing the need to load from the disk after training)

## PR review    

### Test Plan
To verify the success, run the Hubert Pretraining pipeline, load a checkpoint model, and verify that the "dictionaries" key is present in the state within the model.

Specifically, 
```
PYTHONPATH=. python /path/to/fairseq/fairseq_cli/hydra_train.py -m \
        --config-dir ${fairseq_dir}/examples/hubert/config/pretrain \
        --config-name hubert_base_librispeech \
        hydra/launcher=submitit_local \
        hydra.launcher.gpus_per_node=2 \
        hydra.launcher.cpus_per_task=8 \
        hydra.launcher.mem_gb=384 \
        task.data=${tsv_dir} \
        task.label_dir=${km_dir} \
        task.labels=["km"] \
        +data=iter1 \
        optimization.max_update=250 \
        hydra.sweep.dir=${exp_dir} \
        hydra.run.dir=${exp_dir} > ${exp_dir}/log.out 2> ${exp_dir}/log.err &
```
Then, at the location of the model, load the model using `pytorch.load`, and verifying that "dictionaries" is a key under the `task_state` key of the model.

## Did you have fun?

### DIFF
--- a/fairseq/tasks/hubert_pretraining.py
+++ b/fairseq/tasks/hubert_pretraining.py
@@ -107,19 +107,19 @@ class HubertPretrainingTask(FairseqTask):
     def __init__(
         self,
         cfg: HubertPretrainingConfig,
-        dictionaries: Dict[str, Dictionary],
     ) -> None:
         super().__init__(cfg)
 
         logger.info(f"current directory is {os.getcwd()}")
         logger.info(f"HubertPretrainingTask Config {cfg}")
-        self._dictionaries = dictionaries
+        
+        self.state.add_factory("dictionaries", lambda: self.dictionaries_factory(cfg))
+        self.state.add_factory("target_dictionary", self.target_dictionary_factory)
 
+        self._dictionaries = self.state.dictionaries
+        self._target_dictionary = self.state.target_dictionary
         self._source_dictionary = None
-        self._target_dictionary = None
 
-        if len(self.dictionaries) == 1:
-            self._target_dictionary = self.dictionaries[0]
         self.blank_symbol = "<s>"
 
     @property
@@ -138,14 +138,22 @@ class HubertPretrainingTask(FairseqTask):
     def setup_task(
         cls, cfg: HubertPretrainingConfig, **kwargs
     ) -> "HubertPretrainingTask":
+        return cls(cfg)
+        
+    def dictionaries_factory(self, cfg: HubertPretrainingConfig):
         label_dir = cfg.data if cfg.label_dir is None else cfg.label_dir
-        dictionaries = {
+        return {
             label: Dictionary.load(f"{label_dir}/dict.{label}.txt")
             if os.path.exists(f"{label_dir}/dict.{label}.txt")
             else None
             for label in cfg.labels
         }
-        return cls(cfg, dictionaries)
+
+    def target_dictionary_factory(self):
+        if len(self._dictionaries) == 1:
+            logger.info(self._dictionaries)
+            return self._dictionaries[list(self._dictionaries)[0]]
+        return None
 
     def get_label_dir(self) -> str:
         if self.cfg.label_dir is None:
--- a/fairseq/tasks/hubert_pretraining.py
+++ b/fairseq/tasks/hubert_pretraining.py
@@ -115,12 +115,13 @@ class HubertPretrainingTask(FairseqTask):
         logger.info(f"current directory is {os.getcwd()}")
         logger.info(f"HubertPretrainingTask Config {cfg}")
 
+        self.cfg = cfg
         self.fine_tuning = cfg.fine_tuning
 
         if cfg.fine_tuning:
-            self.state.add_factory("target_dictionary", lambda: self.load_dictionaries(cfg)[0])
+            self.state.add_factory("target_dictionary", lambda: self.load_dictionaries)
         else:
-            self.state.add_factory("dictionaries", lambda: self.load_dictionaries(cfg))
+            self.state.add_factory("dictionaries", lambda: self.load_dictionaries)
 
         self._source_dictionary = None
 
@@ -144,9 +145,10 @@ class HubertPretrainingTask(FairseqTask):
     ) -> "HubertPretrainingTask":
         return cls(cfg)
 
-    def load_dictionaries(self, cfg: HubertPretrainingConfig):
-        label_dir = cfg.data if cfg.label_dir is None else cfg.label_dir
-        return [Dictionary.load(f"{label_dir}/dict.{label}.txt") for label in cfg.labels]
+    def load_dictionaries(self):
+        label_dir = self.cfg.data if self.cfg.label_dir is None else self.cfg.label_dir
+        dictionaries = return [Dictionary.load(f"{label_dir}/dict.{label}.txt") for label in self.cfg.labels]
+        return dictionaries[0] if self.cfg.fine_tuning else dictionaries
 
     def get_label_dir(self) -> str:
         if self.cfg.label_dir is None:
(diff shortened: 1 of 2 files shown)


fairseq/tasks/hubert_pretraining.py
```diff
@@ -107,19 +107,19 @@ class HubertPretrainingTask(FairseqTask):
...
 
         logger.info(f"current directory is {os.getcwd()}")
         logger.info(f"HubertPretrainingTask Config {cfg}")
-        self._dictionaries = dictionaries
+        
+        self.state.add_factory("dictionaries", lambda: self.dictionaries_factory(cfg))
+        self.state.add_factory("target_dictionary", self.target_dictionary_factory)
 
+        self._dictionaries = self.state.dictionaries
+        self._target_dictionary = self.state.target_dictionary
```
hikushalhere, line 120: `self._target_dictionary` can be retrieved from `self._dictionaries`. (Lines 121-122 on the left).

fairseq/tasks/hubert_pretraining.py
```diff
@@ -107,19 +107,19 @@ class HubertPretrainingTask(FairseqTask):
...
-        dictionaries: Dict[str, Dictionary],
     ) -> None:
         super().__init__(cfg)
 
         logger.info(f"current directory is {os.getcwd()}")
         logger.info(f"HubertPretrainingTask Config {cfg}")
-        self._dictionaries = dictionaries
+        
+        self.state.add_factory("dictionaries", lambda: self.dictionaries_factory(cfg))
+        self.state.add_factory("target_dictionary", self.target_dictionary_factory)
```
hikushalhere, line 117: Since target dictionary can be retrieved from dictionaries, this is not required.

fairseq/tasks/hubert_pretraining.py
```diff
@@ -138,14 +138,22 @@ def dictionaries(self) -> List[Dictionary]:
...
             else None
             for label in cfg.labels
         }
-        return cls(cfg, dictionaries)
+
+    def target_dictionary_factory(self):
+        if len(self._dictionaries) == 1:
+            logger.info(self._dictionaries)
+            return self._dictionaries[list(self._dictionaries)[0]]
+        return None
```
hikushalhere, line 156: May be returning {} is better to not crash the process, and signal that the target dictionary is empty.

hikushalhere (commented):
Looks good to me. Can you pls add how you tested this? You can add it in "What does this PR do" with a prefix "Test Plan:".

hikushalhere (approved):

fairseq/tasks/hubert_pretraining.py
```diff
@@ -107,19 +109,29 @@ class HubertPretrainingTask(FairseqTask):
...
+            self._dictionaries = self.state.ft_dictionaries
+        else:
+            self.state.add_factory("pt_dictionaries", lambda: self.dictionaries_factory(cfg))
+            self._dictionaries = self.state.pt_dictionaries
+
+        if len(self._dictionaries) == 1:
+            self._target_dictionary = self._dictionaries[list(self._dictionaries)[0]]
+        else:
+            logger.info("Multiple Dictionaries, cannot pick single target.")
+            self._target_dictionary = {}
```
hikushalhere, line 131: @wnhsu How should multiple dictionaries be handled?

fairseq/tasks/hubert_pretraining.py
```diff
@@ -134,18 +146,32 @@ def target_dictionary(self) -> Optional[Dictionary]:
...
         return [self._dictionaries[l] for l in self.cfg.labels]
 
+    @property
+    def pt_dictionaries(self) -> Optional[Dictionary]:
+        if self.fine_tuning:
+            logger.warning("Attempting to access pre-training dictionary during fine-tuning")
+        return self.state.pt_dictionaries
+
+    @property
+    def ft_dictioanries(self) -> Optional[Dictionary]:
```
hikushalhere, line 156: If you intend to return the dictionaries via the new properties `ft_dictionaries` and `pt_dictionaries` then you need to make sure that all callers of the property `dictionaries` are also changed.

However, I don't think you need two properties `ft_dictionaries` and `pt_dictionaries`. You can return the appropriate dictionary from the dictionaries property on line 146 based on `self.fine_tuning`.

hikushalhere (changes requested):

fairseq/tasks/hubert_pretraining.py
neeyanthkvk: This is how the `_target_dictionary` field was handled originally as well.

wnhsu: the logic is a bit convolving when I first design this... so some basics:
1) fine-tuning uses only **one** dictionary
2) pre-training may use **multiple** dictionaries
3) fine-tuning builds the model with `task.target_dictionary`
4) pre-training builds the model with `task.dictionaries`
5) previously we have `task.target_dictionary = task.dictionaries[0]`


I'm going to propose that:
a) we change `HubertModel.build_model` to use `pt_dictionaries` instead of `dictionaries` for clarity
b) rename `dictionaries_factory` to `load_dictionaries`, and return `[Dictionary.load(f"{label_dir}/dict.{label}.txt") for label in cfg.labels]`. Note that we do not need to check if the file exists anymore since it was a workaround when we didn't save dictionaries into task.state before.
c) we do 
```
if cfg.fine_tuning:
  self.state_add_factory("target_dictionary", lambda: self.load_dictionaries(cfg)[0])
else:
  self.state_add_factory("dictionaries", lambda: self.load_dictionaries(cfg))
```
and remove L127-131. Note that now we use `@property` anymore for simplicity

Does that sound good?

wnhsu (changes requested):

wnhsu (approved):

fairseq/tasks/hubert_pretraining.py
```diff
@@ -107,19 +109,21 @@ class HubertPretrainingTask(FairseqTask):
...
         logger.info(f"current directory is {os.getcwd()}")
         logger.info(f"HubertPretrainingTask Config {cfg}")
-        self._dictionaries = dictionaries
+
+        self.fine_tuning = cfg.fine_tuning
+
+        if cfg.fine_tuning:
+            self.state.add_factory("target_dictionary", lambda: self.load_dictionaries(cfg)[0])
+        else:
+            self.state.add_factory("dictionaries", lambda: self.load_dictionaries(cfg))
```
hikushalhere, line 123: What will happen here is that sef.load_dictionaries(cfg) will be called right away and the function you're passing to add_factory() is an anonymous function that returns the stored object. May be you can save cfg as member variable (self.cfg = cfg) before calling the function. So then you can pass just the function object.

Alternatively, you can use functools.partial to create a function object with the argument that you can pass to add_factory().

fairseq/tasks/hubert_pretraining.py
```diff
@@ -128,24 +132,21 @@ def source_dictionary(self) -> Optional[Dictionary]:
 
     @property
     def target_dictionary(self) -> Optional[Dictionary]:
-        return self._target_dictionary
+        return self.state.target_dictionary
 
     @property
     def dictionaries(self) -> List[Dictionary]:
-        return [self._dictionaries[l] for l in self.cfg.labels]
+        return self.state.dictionaries
```
hikushalhere, line 139: I think this needs to be: `return self.state.dictionaries if not self.fine_tuning else [self.state.target_dictionary]`

wnhsu: do we need that? I think `dictionaries` is only used for pre-training; whereas `target_dictionary` is used for fine-tuning

Issue #3010, opened 2020-12-09
Problem with exporting wav2vec2 to onnx


voronc:
I am trying to export wav2vec2 model to onnx format, but have an issue.
I inserted the following code to infer.py

wav_input_16khz = torch.randn(1,10000)
torch.onnx.export(models[0], wav_input_16khz, "fairseq_transformer.onnx", verbose=True, operator_export_type=torch.onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)

And got an issue:

`  File "examples/speech_recognition/infer.py", line 443, in <module>
    cli_main()
  File "examples/speech_recognition/infer.py", line 439, in cli_main
    main(args)
  File "examples/speech_recognition/infer.py", line 271, in main
    torch.onnx.export(models[0], wav_input_16khz, "fairseq_transformer.onnx", verbose=True, operator_export_type=torch.onnx.OperatorExportTypes.ONNX_ATEN_FALLBACK)
  File "/usr/local/lib/python3.8/site-packages/torch/onnx/__init__.py", line 225, in export
    return utils.export(model, args, f, export_params, verbose, training,
  File "/usr/local/lib/python3.8/site-packages/torch/onnx/utils.py", line 85, in export
    _export(model, args, f, export_params, verbose, training, input_names, output_names,
  File "/usr/local/lib/python3.8/site-packages/torch/onnx/utils.py", line 632, in _export
    _model_to_graph(model, args, verbose, input_names,
  File "/usr/local/lib/python3.8/site-packages/torch/onnx/utils.py", line 409, in _model_to_graph
    graph, params, torch_out = _create_jit_graph(model, args,
  File "/usr/local/lib/python3.8/site-packages/torch/onnx/utils.py", line 379, in _create_jit_graph
    graph, torch_out = _trace_and_get_graph_from_model(model, args)
  File "/usr/local/lib/python3.8/site-packages/torch/onnx/utils.py", line 342, in _trace_and_get_graph_from_model
    torch.jit._get_trace_graph(model, args, strict=False, _force_outplace=False, _return_inputs_states=True)
  File "/usr/local/lib/python3.8/site-packages/torch/jit/_trace.py", line 1148, in _get_trace_graph
    outs = ONNXTracedModule(f, strict, _force_outplace, return_inputs, _return_inputs_states)(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/torch/nn/modules/module.py", line 727, in _call_impl
[...]

PR #3609, opened 2021-06-10
Add deployment support wav2vec2.0 via torchaudio

mthrok:
## What does this PR do?

In upcoming PyTorch 1.9 / torchaudio 0.9 release, torchaudio support TorchScript-able wav2vec2.0 model definitions.
This PR adds the illustration fo how to convert the models to from `fairseq` and `transformers` into deployable package.

cc @myleott @alexeib

### DIFF
--- a/examples/wav2vec/README.md
+++ b/examples/wav2vec/README.md
@@ -203,6 +203,48 @@ loss = model(input_values, labels=labels).loss
 loss.backward()
 ```
 
+## Deploying wav2vec 2.0 with torchaudio
+
+`torchaudio` has added wav2vec 2.0 model definition that supports TorchScript, alongside of function to import model instances from `fairseq` or 🤗Transformers. By using TorchScript, you can deploy your wav2vec 2.0 model to ONNX runtime, [C++](https://github.com/pytorch/audio/tree/master/examples/libtorchaudio/speech_recognition), [iOS](https://github.com/pytorch/ios-demo-app/tree/master/SpeechRecognition), and [Android](https://github.com/pytorch/android-demo-app/tree/master/SpeechRecognition).
+
+The following code snipet illustrates the process to import a model and convert it to deployable TorchScript object.
+
+### Import a fine-tuned model from `fairseq`
+
+```python
+import fairseq
+from torchaudio.models.wav2vec2.utils import import_fairseq_model
+
+original, _, _ = fairseq.checkpoint_utils.load_model_ensemble_and_task(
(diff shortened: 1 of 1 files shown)


examples/wav2vec/README.md
```diff
@@ -199,6 +199,48 @@ loss = model(input_values, labels=labels).loss
 loss.backward()
 ```
 
+## Deploying wav2vec 2.0 with torchaudio
+
+`torchaudio` has added wav2vec 2.0 model definition that supports TorchScript, alongside of function to import model instances from `fairseq` or 🤗Transformers. By using TorchScript, you can deploy your wav2vec 2.0 model to ONNX runtime, [C++](https://github.com/pytorch/audio/tree/master/examples/libtorchaudio/speech_recognition), [iOS](https://github.com/pytorch/ios-demo-app/tree/master/SpeechRecognition), and [Android](https://github.com/pytorch/android-demo-app/tree/master/SpeechRecognition).
```
mthrok, line 204: Examples for iOS and Android are being updated to use torchaudio.

- https://github.com/pytorch/android-demo-app/pull/141
- https://github.com/pytorch/ios-demo-app/pull/53

ONNX support via TorchScript is reported to work here
- https://github.com/pytorch/fairseq/issues/3010

examples/wav2vec/README.md
```diff
@@ -199,6 +199,48 @@ loss = model(input_values, labels=labels).loss
...
+The following code snipet illustrates the process to import a model and convert it to deployable TorchScript object.
+
+### Import a fine-tuned model from `fairseq`
+
+```python
+import fairseq
+from torchaudio.models.wav2vec2.utils import import_fairseq_model
+
+original, _, _ = fairseq.checkpoint_utils.load_model_ensemble_and_task(
+    ["wav2vec_small_960h.pt"], arg_overrides={'data': "<DIRECTORY_WITH_DICTIONARY>"})
```
alexeib, line 215: in newer models, dictionary is actually stored inside the model checkpoint rather than internally. we could probably convert older checkpoints to follow this new format if it will simplify things

alexeib (commented):
looks great!

examples/wav2vec/README.md
mthrok: @alexeib I removed the `arg_overrides`. Did you update the public checkpoints? My concern is that users who just want to try the published model (if not converted to the new format) will encounter an issue when loading models.

alexeib: no, i have not updated the old checkpoints yet. maybe we can allow users to optionally provide data dir instead of making it mandatory to either provide or not?

PR #3659, opened 2021-06-28
Hydra 1.1 compatibility: Use an explicit schema for the primary config

omry:
## What does this PR do?
Fixes compatibility with Hydra 1.1.
The result is compatible with both Hydra 1.0 and Hydra 1.1, and will allow a smoother migration to Hydra 1.1.

At this point I am not yet removing the restriction on the Hydra version from setup.py:
1. It depends on some Hydra 1.1 changes that are not yet released (It will be compatible with 1.1.1).
2. Upgrading will result in deprecation warnings, and fixing them will break compatibility with Hydra 1.0.

There will be some followup to make the code fully compatible with 1.1 once Hydra 1.1 is the default version in fbcode.

### DIFF
--- a/fairseq/checkpoint_utils.py
+++ b/fairseq/checkpoint_utils.py
@@ -203,7 +203,7 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
         cfg.restore_file == "checkpoint_last.pt"
     ):  # default value of restore_file is 'checkpoint_last.pt'
         checkpoint_path = os.path.join(
-            cfg.save_dir, "checkpoint_last{}.pt".format(suffix)
+            cfg.get("save_dir"), "checkpoint_last{}.pt".format(suffix)
         )
         first_launch = not PathManager.exists(checkpoint_path)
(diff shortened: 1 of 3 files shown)


fairseq/checkpoint_utils.py
```diff
@@ -203,7 +203,7 @@ def load_checkpoint(cfg: CheckpointConfig, trainer, **passthrough_args):
         cfg.restore_file == "checkpoint_last.pt"
     ):  # default value of restore_file is 'checkpoint_last.pt'
         checkpoint_path = os.path.join(
-            cfg.save_dir, "checkpoint_last{}.pt".format(suffix)
+            cfg.get("save_dir"), "checkpoint_last{}.pt".format(suffix)
```
omry, line 206: That's a weird one:
save_dir is not in the config during the test.
In Hydra 1.0, this returned None.
In Hydra 1.1, this raises an error.

os.path.join normally would fail if it gets None, but it passes because the test is mocking it.
The fix here is just retaining the existing weird behavior (join with a None save_dir).

PR #3660, opened 2021-06-28
removed numpy version restriction on Python 3.6

omry:
This is conflicting with a recent change on fairscale:

https://github.com/facebookresearch/fairscale/pull/731


omry:
As of numpy 1.20, the minimum supported Python version is 3.7.
Closing, there is a discussion to have about fairseq minimum Python version.

PR #3722, opened 2021-07-19
Compatibility fix with Hydra 1.1

omry:
One of the changes in Hydra 1.1 is that the default composition order is changing.
This is documented [here](https://hydra.cc/docs/upgrades/1.0_to_1.1/default_composition_order).
In Hydra 1.1, a config is overriding values introduced by the defaults list while in Hydra 1.0 - the defaults list was overriding the values in the config.

fairseq is currently depending on the previous behavior:
The class `FairseqConfig` defines config values, and it's expecting them to be overridden by the defaults list.
This result in a different config being created when running `fairseq_cli/hydra_train.py` with Hydra 1.0 and with 1.1.

Hydra 1.1 introduced the `_self_` keyword in the defaults list to control the composition order. In order to achieve the behavior of Hydra 1.0, `_self_` should be added as the first item in the defaults list.

To allow for a smoother migration, Hydra 1.0 is ignoring `_self_` starting from 1.0.7 (previous versions will issue an error).

This diff adds `_self_` as the first item in the defaults list the fairseq config, and introduce a dependency a Hydra 1.0 version that is equal or newer to 1.0.7.

### Testing:
I ensured that the following yield the same composed config:
Default config with Hydra 1.0.6, 1.0.7 and 1.1.0

`examples/wav2vec/config/finetuning/base_10h.yaml` with Hydra 1.0.6, 1.0.7 and 1.1.0.

This can be achieved by outputing the generated config using `--cfg job` and compating the outputs.


fairseq/optim/fp16_optimizer.py
```diff
@@ -479,7 +479,7 @@ def __init__(
                 "Unsupported optimizer: {}".format(optimizer.__class__.__name__)
             )
 
-        super().__init__(cfg.optimizer)
+        super().__init__(getattr(cfg, "optimizer", None))
```
omry, line 482: This fails in some tests on Hydra 1.1 due to an OmegaConf 2.1 change making access to non existing fields behave like it does for objects and dicts (raises an exception).

alexeib:
will this have any effect on external configs (--config-dir + --config-name)? if not lgtm

omry:
Not supposed to.
As I said, I tested examples/wav2vec/config/finetuning/base_10h.yaml and it's composing the same config.

omry:
@alexeib, can we get this merged soon?

alexeib:
sure, can you just make diff from this? merging happens through fbcode, not here

gegallego:
Hi! [This line](https://github.com/facebookresearch/fairseq/blob/acd9a53607d1e5c64604e88fc9601d0ee56fd6f1/setup.py#L203) could be changed to `hydra-core>=1.0.7,<=1.1`, right?
Also, `omegaconf` requirement should then be changed to `omegaconf<=2.1` to satisfy [this requirement](https://github.com/facebookresearch/hydra/blob/79b39bed17eadae1ea98d30c520c4686fedab530/requirements/requirements.txt#L1) from `hydra`.

rsxdalv:
@omry Could you please confirm if fairseq works with hydra-core==1.1?

omry:
@rsxdalv, Nope. I am no longer involved with either.

PR #3742, opened 2021-07-27
Fixed the reference to mask_channel_prob in task cfg

ultrons:
Updated example config file for tpu to include mask parameters.
Noted currently cli bug in README

## What does this PR do?
Fixes # (3741) B.

### DIFF
--- a/fairseq/tasks/audio_pretraining.py
+++ b/fairseq/tasks/audio_pretraining.py
@@ -242,7 +242,7 @@ class AudioPretrainingTask(FairseqTask):
                 **self._get_mask_precompute_kwargs(task_cfg),
             )
 
-        if self.cfg.tpu and task_cfg["mask_channel_prob"] == 0.0:
+        if self.cfg.tpu and task_cfg.inferred_w2v_config.mask_channel_prob == 0.0:
             logger.info(
                 "Pretraining on TPUs may suffer convergence "
                 "issues when training with `mask_channel_prob` value of "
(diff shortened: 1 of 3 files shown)


fairseq/tasks/audio_pretraining.py
```diff
@@ -140,8 +140,8 @@ class AudioPretrainingConfig(FairseqDataclass):
         },
     )
 
-    inferred_w2v_config: Optional[InferredW2vConfig] = field(
-        default=None,
```
alexeib, line 144: do we still need to change this? can we keep it as before?

fairseq/tasks/audio_pretraining.py
```diff
@@ -242,7 +242,7 @@ def load_dataset(self, split: str, task_cfg: FairseqDataclass = None, **kwargs):
                 **self._get_mask_precompute_kwargs(task_cfg),
             )
 
-        if self.cfg.tpu and task_cfg["mask_channel_prob"] == 0.0:
+        if self.cfg.tpu and task_cfg.inferred_w2v_config.mask_channel_prob == 0.0:
```
alexeib, line 245: 👍

alexeib (changes requested):

fairseq/tasks/audio_pretraining.py
ultrons: It was one of the intermediate changes we did when trying to solve the nested dataclass issue. The right answer is I don't know. Do you recommend to revert it.?

ultrons: Changed back to orginal optional and default value.

PR #3873, opened 2021-09-14
Add Tutel boost for Fairseq MoE acceleration

ghostplant:
Add optimized library implementation to accelerator MoE module.

# Before submitting

## What does this PR do?
To accelerate MoE execution in fairseq if tutel package is installed.

## PR review    

## Did you have fun?

### DIFF
--- a/fairseq/modules/moe/moe_layer.py
+++ b/fairseq/modules/moe/moe_layer.py
@@ -24,7 +24,7 @@ else:
 
 try:
     # To enable Tutel MoE optimizations:
-    #   python3 -m pip install --user https://github.com/microsoft/tutel/releases/download/v0.1.0/tutel-0.1.0.tar.gz
+    #   python3 -m pip install --user --upgrade git+https://github.com/microsoft/tutel@v0.1.x
     from tutel import moe as tutel_moe
 
     has_tutel, fused_cumsum_sub_one = True, tutel_moe.fast_cumsum_sub_one
@@ -167,7 +167,7 @@ class MOELayer(Base):
 
             if not hasattr(self, '_tutel_dispatcher'):
                 self._tutel_dispatcher = tutel_moe.fast_dispatcher(E, C, M, dispatch_dtype=reshaped_input.dtype)
-            self._tutel_dispatcher.update(indices_, locations_, gates_)
+            self._tutel_dispatcher.update(indices_, locations_, gates_, capacity=C)
             dispatched_input = self._tutel_dispatcher.encode(reshaped_input)
         else:
             l_aux, combine_weights, dispatch_mask, self.metadata = self.gate(reshaped_input, reshaped_input_padding_mask)
--- a/fairseq/modules/moe/moe_layer.py
+++ b/fairseq/modules/moe/moe_layer.py
@@ -22,6 +22,14 @@ if TYPE_CHECKING:
 else:
     Base = Module
 
+try:
+    # To enable Tutel MoE optimizations:
+    #   python3 -m pip install --user https://github.com/microsoft/tutel/releases/download/v0.1.0/tutel-0.1.0.tar.gz
+    from tutel import moe as tutel_moe
+
+    has_tutel, fused_cumsum_sub_one = True, tutel_moe.fast_cumsum_sub_one
+except ModuleNotFoundError:
+    has_tutel, fused_cumsum_sub_one = False, lambda mask: torch.cumsum(mask, dim=0) - 1
 
 logger = logging.getLogger(__name__)
 
@@ -153,14 +161,23 @@ class MOELayer(Base):
                 padded_input_padding_mask[:reshaped_input_shape[0]] = False
             reshaped_input_padding_mask = padded_input_padding_mask
 
-        l_aux, combine_weights, dispatch_mask, self.metadata = self.gate(reshaped_input, reshaped_input_padding_mask)
+        if has_tutel:
+            l_aux, self.metadata, C, E, indices_, locations_, gates_ = self.gate(reshaped_input, reshaped_input_padding_mask)
+            S, M = reshaped_input.size(0), reshaped_input.size(1)
 
-        dispatch_mask = dispatch_mask.to(input.dtype).permute(1, 2, 0)  # S,E,C -> E,C,S
-        E, C, S = dispatch_mask.size()
-        M = reshaped_input.size(1)
-        assert reshaped_input.size() == (S, M)
-        # einsum("sec,sm->ecm")
-        dispatched_input = torch.mm(dispatch_mask.view(E*C, S), reshaped_input)  # -> (E*C),M
+            if not hasattr(self, '_tutel_dispatcher'):
+                self._tutel_dispatcher = tutel_moe.fast_dispatcher(E, C, M, dispatch_dtype=reshaped_input.dtype)
+            self._tutel_dispatcher.update(indices_, locations_, gates_)
+            dispatched_input = self._tutel_dispatcher.encode(reshaped_input)
+        else:
+            l_aux, combine_weights, dispatch_mask, self.metadata = self.gate(reshaped_input, reshaped_input_padding_mask)
+
+            dispatch_mask = dispatch_mask.to(input.dtype).permute(1, 2, 0)  # S,E,C -> E,C,S
+            E, C, S = dispatch_mask.size()
+            M = reshaped_input.size(1)
+            assert reshaped_input.size() == (S, M)
+            # einsum("sec,sm->ecm")
+            dispatched_input = torch.mm(dispatch_mask.view(E*C, S), reshaped_input)  # -> (E*C),M
 
         if self.all2all_size > 1:
             dispatched_input = self.all_to_all_wrapper(dispatched_input)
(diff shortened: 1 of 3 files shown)


ngoyal2707:
This seems to be breaking when running on multiple processes (gpus).

```
  File "/private/home/namangoyal/src/fairseq_gshard/fairseq-py/fairseq/modules/moe/moe_layer.py", line 165, in forward
    l_aux, self.metadata, C, E, indices_, locations_, gates_ = self.gate(reshaped_input, reshaped_input_padding_mask)
  File "/private/home/namangoyal/.conda/envs/fairseq-20210102-py38/lib/python3.8/site-packages/torch/nn/modules/module.py", line 727, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/private/home/namangoyal/src/fairseq_gshard/fairseq-py/fairseq/modules/moe/top2gate.py", line 247, in forward
    return top2gating(
  File "/private/home/namangoyal/src/fairseq_gshard/fairseq-py/fairseq/modules/moe/top2gate.py", line 131, in top2gating
    locations1 = fused_cumsum_sub_one(mask1)
  File "/private/home/namangoyal/src/tutel/tutel/jit_kernels/gating.py", line 79, in fast_cumsum_sub_one
    return get_cumsum_kernel(data.size(0), data.size(1))(data)
  File "/private/home/namangoyal/src/tutel/tutel/jit_kernels/gating.py", line 70, in optimized_cumsum
    base_kernel(mask1.to(torch.int32), locations1)
  File "/private/home/namangoyal/src/tutel/tutel/impls/jit_compiler.py", line 44, in func
    tutel_custom_kernel.invoke(inputs, __ctx__ * 256 + key)
RuntimeError: (0) == (return_code) INTERNAL ASSERT FAILED at "/private/home/namangoyal/src/tutel/tutel/custom/custom_kernel.cpp":95, please report a bug to PyTorch. return_codeCHECK_EQ fails.
```

I think, this could be due to some weird interaction of how Tutel compiles and loads custom cuda kernels and maybe spawned processes.  

I tried Tutel's example also on multi gpus with pytorch distribuetd launch script, that seems to work. One difference I think is, pytorch launch script forks the new processes vs fairseq spawns them. Maybe thats creating a problem?
I also created a `spawn version` of Tutel's helloworld example, and even that doesnt work. I can share that if that can be helpful in debugging?

ghostplant:
Thanks, @ngoyal2707 for reporting this issue. From tutel's team, we also use pytorch distribuetd launch to test this pull requests, and they are verified running well using both pytorch's ddp backend and fully-sharded backend.

So mostly likely it is an environmental issue. I see the assertion failure you encountered happens at: https://github.com/microsoft/tutel/blob/v0.1.x/tutel/custom/custom_kernel.cpp#L95
It indicates that the JIT compiler doesn't get a source code content that matches its expected format.

May I ask whether you change some source codes and install Tutel from source? If so, the changes may be not suitable and break the following procedure.

If you don't change anything or just install tutel via `pip install https://..`, can you help to show the standard output of the following shell commands in your environment?
```sh
(Command-1)
$ ls /tmp/*.cu

(Command-2)
$ cat /tmp/0-0.cu
```

ngoyal2707:
1) pytorch distributed launch script works for me also, I am saying that fairseq's way of launching doesn't work for me.
Can you please share the cmd you used to test fairseq integration? 

2) The aassertion fails here for me:
```
https://github.com/microsoft/tutel/blob/v0.1.x/tutel/custom/custom_kernel.cpp#L92 
```
the trace says 95 cause I have bunch of debugging print statements in that code locally.

3) I tried installing without any changes and it still failed.

4) I do see all the *.cu files and the fatbins also in the tmp dir, they just dont load with spawned processes:
````
/tmp/0-0.cu  /tmp/1-0.cu  /tmp/10-0.cu  /tmp/12-0.cu  /tmp/14-0.cu  /tmp/16-0.cu  /tmp/18-0.cu  /tmp/2-0.cu  /tmp/20-0.cu  /tmp/22-0.cu  /tmp/24-0.cu  /tmp/26-0.cu  /tmp/28-0.cu  /tmp/3-0.cu  /tmp/30-0.cu  /tmp/32-0.cu  /tmp/34-0.cu  /tmp/36-0.cu  /tmp/5-0.cu  /tmp/7-0.cu  /tmp/9-0.cu
/tmp/0-1.cu  /tmp/1-1.cu  /tmp/11-0.cu  /tmp/13-0.cu  /tmp/15-0.cu  /tmp/17-0.cu  /tmp/19-0.cu  /tmp/2-1.cu  /tmp/21-0.cu  /tmp/23-0.cu  /tmp/25-0.cu  /tmp/27-0.cu  /tmp/29-0.cu  /tmp/3-1.cu  /tmp/31-0.cu  /tmp/33-0.cu  /tmp/35-0.cu  /tmp/4-0.cu   /tmp/6-0.cu 
```

ghostplant:
@ngoyal2707 The outputs make sense.
The format of these JIT codes are: `<code_id>-<gpu-id>.cu`, and they are created with different names to avoid being overwritten by different child processes.

Seems like the fairseq's launch doesn't deal with `LOCAL_RANK` at https://github.com/microsoft/tutel/blob/v0.1.x/tutel/impls/jit_compiler.py#L20.

So multiple child executors from fairseq's launch are conflicts writing into the same file target.
Can you have a quick try on change that line into random values for different executors to avoid the conflict? Those values would also be smaller than 256.
If that works, the root cause is found, and I'll think about how to fix it in a gentle way.

ngoyal2707:
Ahhh perfect, now I got it! thanks

ghostplant:
@ngoyal2707 BTW, may I ask what command you use to launch the model using **fairseq's way**? We want to see the runtime differences of that way and it would help us to come up with a gentle fix. Thanks!

jspark1105:
@ghostplant thanks for the awesome stuffs in Tutel. Curious if you considered using NVRTC instead of invoking nvcc. I'm asking because in some environment with VMs and so on don't have access to run nvcc. I quickly tried and functions in gating .py can be compiled with NVRTC with minor tweak but the ones in sparse.py have issues associated with __half2 doesn't seem to be supported by NVRTC. Wonder if you have any experience.

ghostplant:
@jspark1105 A good suggestion. In fact, our first version was based on NVRTC initially, but we soon find that **same kernels compiled by NVRTC is less optimized than using NVCC**. NVRTC's optimization options that NVIDIA exposes are very limited.

We create a temporary branch named `jit_with_nvrtc`: https://github.com/microsoft/tutel/tree/jit_with_nvrtc
If you are interested, you can install it from source and see whether the hello-world example is a little slower. (about 3% slower?)

jspark1105:
Thanks so much @ghostplant. I tried the branch but not passing a simple correctness check I wrote. If I comment out "#define USE_NVRTC" it goes through. I'll dig more and share minimum reproducer but wonder if you have any guess.

ghostplant:
@jspisak Not quite sure, Is your check script compatible with NVRTC?

ghostplant:
@jspark1105 One possible guess: can you remove some fast-but-low-precision compiler option at: https://github.com/microsoft/tutel/blob/v0.1.x/tutel/custom/custom_kernel.cpp#L61

Like `--use_fast_math`, which is to use faster instruction to get similar but not precise result.

fairseq/modules/moe/moe_layer.py
```diff
@@ -153,14 +161,23 @@ def forward(self, *input: Tensor, input_padding_mask=None, **kwargs: Any) -> Ten
                 padded_input_padding_mask[:reshaped_input_shape[0]] = False
             reshaped_input_padding_mask = padded_input_padding_mask
 
-        l_aux, combine_weights, dispatch_mask, self.metadata = self.gate(reshaped_input, reshaped_input_padding_mask)
+        if has_tutel:
+            l_aux, self.metadata, C, E, indices_, locations_, gates_ = self.gate(reshaped_input, reshaped_input_padding_mask)
+            S, M = reshaped_input.size(0), reshaped_input.size(1)
+
+            if not hasattr(self, '_tutel_dispatcher'):
+                self._tutel_dispatcher = tutel_moe.fast_dispatcher(E, C, M, dispatch_dtype=reshaped_input.dtype)
```
vedanuj, line 169: Since we are initializing the `fast_dispatcher` with `E`, `C`, `M` values from the first forward, this will create issue when the capacity `C` changes (either when number of tokens are not equal to max_tokens or during eval). In the `.update` call, we should also update the `self.capacity` in the  `fast_dispatcher` for every forward.

cc @ngoyal2707

ghostplant: Thanks for this requirement. The JIT kernels have to handle dynamic size C to satisfy that, while current version only consider S (batch * tokens) as dynamic values. @ngoyal2707 Does your check also target to the same case?

ghostplant: @vedanuj I just append a second commit for this PR to allow `capacity` to be reset when calling `_tutel_dispatcher.update(..)`.

jspark1105:
@ghostplant following up nvrtc, found removing --restrict option from option resolves the issue.

ghostplant:
@jspark1105 Thanks, just a little weird since this keyword is expected to work the same way of `__restrict__` and it works as expected in my environment. Do you know how this NVRTC's `--restrict` option impacts on your check correctness?

PR #3922, opened 2021-09-29
Update README to match master->main branch change

erichan1:
Loading xlmr doesn't work because trying to pull from master branch, which has been changed to main.


dianaml0 (approved):
Thanks!

PR #3989, opened 2021-10-31
Fixes OOM bug causing distributed training jobs to hang.

telmop:
# Before submitting

- [X] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/main/CONTRIBUTING.md)?
- [X] Did you make sure to update the docs?
- [X] Did you write any new necessary tests?

## What does this PR do?
Fixes issue where fairseq training hangs when using multiple workers. When one of the workers gets a OOM error, it tries to recover with a `try ... except` statement, but the other workers keep waiting for it, and they get out of sync.

### DIFF
--- a/fairseq/trainer.py
+++ b/fairseq/trainer.py
@@ -783,6 +783,10 @@ class Trainer(object):
                         torch.cuda.empty_cache()
                     if self.cfg.distributed_training.distributed_world_size == 1:
                         return None
+                    else:
+                        # Re-raise exception, as otherwise distributed workers
+                        # get out of sync and cause the training job to hang.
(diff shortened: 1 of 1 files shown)


tberckmann:
I am hitting the same issue on distributed jobs.

With your change, now my job ends when I hit an OOM. But I thought the correct behavior is to handle the OOM gracefully and continue training. Did I not understand the correct behavior properly?

telmop:
I agree that would be preferable behavior @tberckmann, but I'd rather my jobs terminate than enter an infinite loop. That would be a great follow-up PR.

PR #3998, opened 2021-11-04
enables audio padding rather than cropping for HuBERT finetuning and decoding. Also fixes the issue that the decoding results are non-deterministic due to random audio cropping

freewym:
# Before submitting

## What does this PR do?
Enables audio padding rather than cropping for HuBERT finetuning and decoding (it shouldn't crop the audio during CTC finetuning, otherwise we may lose audio corresponding to some part of the transcript). Also fixes the issue that the decoding results of HuBERT  models are non-deterministic across different runs of inference with exactly the same setting, by disabling random_crop during inference (also disable it for the fine-tuning stage, as there seems no point doing wave crop for fine-tuning, which is also consistent with wav2vec2).

## PR review

## Did you have fun?

### DIFF
--- a/examples/hubert/config/decode/infer_fsqlm.yaml
+++ b/examples/hubert/config/decode/infer_fsqlm.yaml
@@ -16,6 +16,7 @@ task:
   fine_tuning: true
   data: ???
   normalize: ???
+  pad_audio: true
 
 decoding:
   type: fairseqlm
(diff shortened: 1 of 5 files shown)


freewym:
@dianaml0 Could you also please consider this fix?

freewym:
@hirofumi0810 could you please do a review for this?

PR #4174, opened 2022-02-01
Adding scalar quantization aware training support for the hf_gpt2 model

amandeephy:
# Before submitting

- [No] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [Yes] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/main/CONTRIBUTING.md)?
- [Yes, comments around changes and help for the new flag] Did you make sure to update the docs?
- [Yes, the wikitext-103 language modelling example under scalar quant noise can now be used with the `--arch` hf_gpt2 ] Did you write any new necessary tests?

```
BASE_DIR=<BASE_DIR>
MAX_TOKENS=1024
CUDA_VISIBLE_DEVICES=0,1 fairseq-train \
  --task language_modeling $BASE_DIR/data-bin/wikitext-103 \
  --save-dir $BASE_DIR/hf_gpt2_wikitext-103/checkpoints \
  --arch hf_gpt2 \
  --dropout 0.1 \
  --optimizer adam --adam-betas '(0.9, 0.98)' \
  --weight-decay 0.01 --clip-norm 0.0 \
  --lr 0.0005 --lr-scheduler inverse_sqrt \
  --warmup-updates 4000 --warmup-init-lr 1e-07 \
  --tokens-per-sample 512 --sample-break-mode none \
  --max-tokens $MAX_TOKENS --update-freq 16 \
  --fp16 \
  --max-update 5000000 \
  --tensorboard-logdir $BASE_DIR/hf_gpt2_wikitext-103/tensorboard \
  --no-epoch-checkpoints \
  --quant-noise-scalar 0.5
```

Also, as an experiment we tried setting up 2 (almost) identical `transformer_lm` and `hf_gpt2` tiny models and trained them on the same dataset for 100k epochs.
 
We see that the models with noise have slightly worse validation losses than those without noise, and also that they take slightly longer to train: 

tlm_w_wo_noise

hf_gpt2_w_wo_noise

## What does this PR do?
This PR adds support for (scalar) quantization aware training of the `hf_gpt2` model (which is a wrapper around the hugginface `GPT2LMHead` model ).

We add the `--quant-noise-scalar` flag as an argument to the `hf_gpt2` model (as is typically used for scalar noise in fairseq models). The model then gets quantized during the `<fairseq_task>.build_model()` phase.

The quantization replaces (in-place) existing modules with their quantized counterparts. To that end, we introduce a new quantized module `qconv1dhf` which replaces the `conv1d` modules in the hf_gpt2 model (similar to what `qlinear` does for the `nn.Linear` module).

## PR review

## Did you have fun?
Ofcourse !

### DIFF
--- a/fairseq/models/huggingface/hf_gpt2.py
+++ b/fairseq/models/huggingface/hf_gpt2.py
@@ -9,13 +9,13 @@ import sys
 from typing import Dict, List, Optional
 
 import torch
+
 from fairseq.models import (
     FairseqIncrementalDecoder,
     FairseqLanguageModel,
     register_model,
     register_model_architecture,
 )
-
 
 logger = logging.getLogger(__name__)
 
@@ -44,6 +44,14 @@ class HuggingFaceGPT2LanguageModel(FairseqLanguageModel):
         parser.add_argument('--attention-dropout', type=float, metavar='D',
                             help='dropout probability for attention weights')
         # fmt: on
+
+        # Adding argument for scalar quantization noise, same name as in transformer_lm model cfg
+        parser.add_argument(
+            "--quant_noise_scalar",
+            type=float,
+            metavar="D",
+            help="scalar quantization noise and scalar quantization at training time",
+        )
 
     @classmethod
     def build_model(cls, args, task):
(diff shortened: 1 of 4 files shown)


amandeephy:
Bump

Issue #4209, opened 2022-02-12
PadDataset does not have pad_length (multilingual_language_modeling)


afcruzs-ms:
## 🐛 Bug

On [`multilingual_language_modeling.py` the method `build_dataset_for_inference`](https://github.com/pytorch/fairseq/blob/f591cc94caa85098ccf125a4782f91125b6a086d/fairseq/tasks/multilingual_language_modeling.py#L533) creates an instance of `PadDataset` and passes `pad_length = max_seq_len` as an argument. However, on `pad_dataset.py` the [constructor of `PadDataset`](https://github.com/pytorch/fairseq/blob/f591cc94caa85098ccf125a4782f91125b6a086d/fairseq/data/pad_dataset.py#L12) does not have the argument `pad_length`.

### To Reproduce

Steps to reproduce the behavior (**always include the command you ran**):

1. Try to run the example at `fairseq/examples/xglm/`
Stack trace:
```
Traceback (most recent call last):
  File "go.py", line 75, in <module>
    predict = COPA_eval(example["premise"], example["choice1"], example["choice2"])
  File "go.py", line 69, in COPA_eval
    lprob1 = get_logprobs(prompt + "\n" + alternative1).sum()
  File "go.py", line 63, in get_logprobs
    return lm.score(prompt, replace_newlines_with_eos=True)['positional_scores']
  File "/home/ancruzsa/fairseq/fairseq/hub_utils.py", line 139, in score
    return self.score(
  File "/home/ancruzsa/fairseq/fairseq/hub_utils.py", line 153, in score
    for hypos in self.generate(
  File "/home/ancruzsa/fairseq/fairseq/hub_utils.py", line 187, in generate
    for batch in self._build_batches(tokenized_sentences, skip_invalid_size_inputs):
  File "/home/ancruzsa/fairseq/fairseq/hub_utils.py", line 275, in _build_batches
    dataset=self.task.build_dataset_for_inference(tokens, lengths),
  File "/home/ancruzsa/fairseq/fairseq/tasks/multilingual_language_modeling.py", line 533, in build_dataset_for_inference
    "src_tokens": PadDataset(
TypeError: __init__() got an unexpected keyword argument 'pad_length'
```

#### Code sample

Example at `fairseq/examples/xglm/`, however this is also broken. See #4210 to be able to get the stacktrace above.

### Expected behavior

Calling `build_dataset_for_inference` should not fail this way.

### Environment

[...]

Issue #4210, opened 2022-02-12
XGLM Example does not run


afcruzs-ms:
## 🐛 Bug

The XGLM example has several issues: https://github.com/pytorch/fairseq/tree/main/examples/xglm: 1) `fairseq.models.build_model` fails because the `model_type` (transformer_lm_gpt2_big_wide) does not exist on the registered models, 2) the issue in #4209.

### To Reproduce

Steps to reproduce the behavior (**always include the command you ran**):

1. Run the examples on  https://github.com/pytorch/fairseq/tree/main/examples/xglm
2. You'll see an error complaining "_name" does not exist, which comes from the fact the `model` object in the `build_model` function is None. This can be worked around by hardcoding `model_type = 'transformer_lm'`.
3. With the workaround in 2), you'll see the issue on #4209. This can be mitigated by removing the `pad_length` arguments to PadDataset in `build_dataset_for_inference`.

With this I'm able to run, however it's unclear if this hacks guarantee the correctness of the model.

Stack trace for error in 2):
```
Traceback (most recent call last):
  File "go.py", line 55, in <module>
    lm = TransformerLanguageModel.from_pretrained(model_dir, bpe='sentencepiece')
  File "/home/ancruzsa/fairseq/fairseq/models/fairseq_model.py", line 267, in from_pretrained
    x = hub_utils.from_pretrained(
  File "/home/ancruzsa/fairseq/fairseq/hub_utils.py", line 73, in from_pretrained
    models, args, task = checkpoint_utils.load_model_ensemble_and_task(
  File "/home/ancruzsa/fairseq/fairseq/checkpoint_utils.py", line 469, in load_model_ensemble_and_task
    model = task.build_model(cfg.model, from_checkpoint=True)
  File "/home/ancruzsa/fairseq/fairseq/tasks/multilingual_language_modeling.py", line 260, in build_model
    model = super().build_model(args, from_checkpoint)
  File "/home/ancruzsa/fairseq/fairseq/tasks/fairseq_task.py", line 671, in build_model
    model = models.build_model(args, self, from_checkpoint)
  File "/home/ancruzsa/fairseq/fairseq/models/__init__.py", line 102, in build_model
    f"Could not infer model type from {cfg}. "
KeyError: "'_name'"
```

Stack trace for error 3):
```
Traceback (most recent call last):
  File "go.py", line 75, in <module>
    predict = COPA_eval(example["premise"], example["choice1"], example["choice2"])
[...]

PR #4212, opened 2022-02-14
add missing transformer arch and update PadDataset

todpole3:
Fix issue #4209 #4210


afcruzs-ms:
Thanks for the quick PR! I should also mention I had to upgrade these dependencies in my local system to get the XGLM example to work:

- Sentencepiece: 0.1.85 => 0.1.96
- Numpy: 1.20.3 => 1.22.2

Maybe that can be mentioned on the XGLM readme?

PR #4230, opened 2022-02-21
[Data2Vec] loading data2vec checkpoitns for audio

patrickvonplaten:
Currently loading the data2vec audio checkpoints: https://github.com/pytorch/fairseq/tree/main/examples/data2vec#speech results in a missing parameter value of the omega config.

This PR is solves the problem by adding the parameter. Not sure if this is the correct way to deal with it though. 

cc @alexeib

### DIFF
--- a/fairseq/tasks/audio_finetuning.py
+++ b/fairseq/tasks/audio_finetuning.py
@@ -44,6 +44,9 @@ def label_len_fn(label):
 
 @dataclass
 class AudioFinetuningConfig(AudioPretrainingConfig):
+    cache_in_scratch: bool = field(
+        default=False, metadata={"help": "Whether to cache model in scratch."}
+    )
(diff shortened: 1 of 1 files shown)


alexeib:
ahh its because i migrated checkpoints from a different branch that had a bunch of debug options. it is better to remove this option from the checkpoint (new checkpoints wont have it)

patrickvonplaten:
> ahh its because i migrated checkpoints from a different branch that had a bunch of debug options. it is better to remove this option from the checkpoint (new checkpoints wont have it)

We ported all the checkpoints to Hugging Face Transformers' now already (should be up by the end of this week). So this shouldn't delay the process really ;-)

xmz1009:
@alexeib i got the same bug,why loading the data2vec audio checkpoints results in a missing parameter value of the omega config.what can i fix it.

PR #4311, opened 2022-03-26
Add Aim support for logging

arkkln:
# Before submitting

## What does this PR do?
Enables logging of params and metrics with Aim. Aim is an open-source experiment tracker - https://github.com/aimhubio/aim

1. Added two arguments to CommonConfig:
- aim_repo: defines Aim repository location, can be set to remote URL as well(i.e. `aim://<ip>:<port>`)
- aim_run_hash: defines run hash. If skipped, run will be created or continued based on `save_dir` argument. If there is an existing run which has the same `save_dir`, it will be reopened/continued, otherwise a new run will be created.

2. Implemented AimProgressBarWrapper class to handle logging


ArmenAg:
Just for some context here, our internal research project at FAIR needed to use Aim given they had logging functionality that was not available in either Tensorboard or WAndB. I reached out to the Aim team and they very quickly sent out this to PR integrating with fairseq.

fairseq/logging/progress_bar.py
```diff
@@ -310,6 +321,87 @@ def print(self, stats, tag=None, step=None):
...
+            logger.info(f"Storing logs at Aim repo: {aim_repo}")
+
+            if not aim_run_hash:
+                # Find run based on save_dir parameter
+                query = f"run.checkpoint.save_dir == '{aim_param_checkpoint_dir}'"
+                try:
+                    runs_generator = AimRepo(aim_repo).query_runs(query)
+                    run = next(runs_generator.iter_runs())
+                    aim_run_hash = run.run.hash
+                except:
```
ArmenAg: Do we know what possible exceptions might be thrown here? Don't want to accidently catch events like memory errors, interrupts, system exit etc...

arkkln: sure, will resolve this ASAP

alberttorosyan: @ArmenAg the range of exceptions possibly raised here might be too broad. The reason is that `Repo` class is a core of Aim SDK and covers creation/opening of repo, queries and some data collection as well. Also it depends on 3rd party libraries (`rocksdb` and `sqlite`) which can raise their own exceptions as well...
To make sure we don't miss anything, I would suggest to catch and re-raise exceptions from the following list:
`KeyboardInterrupt` `MemoryError` `SystemError` `SystemExit` `OSError`
For the rest, silently catch the exception and continue.
Let me know if this is sound reasonable, so I can change the code accordingly.

ArmenAg: I think if you do `except Exception:` this will catch all exceptions except SystemExit, KeyboardInterrupt and some other core ones (refer to this [link](https://stackoverflow.com/questions/7160983/catching-all-exceptions-in-python)). I think I'd be okay with this.

alberttorosyan: Yes, seems `Exception` is enough generic to cover Aim and used 3rd parties. Thanks!

alberttorosyan: Done ✅

PR #4480, opened 2022-06-12
BT enablement on fairseq - fairseq change

frank-wei:
Summary:
as titled and depends on D36057338
Fork the inference path inside the forward function. If loaded the checkpoint file and perform the inference, we will deploy BT. Otherwise, fairseq take the position.

In summary:
Accuracy: accuracy loss due to the fp16, the maximum diff is around 0.009. If we set it to fp32, there is no accuracy loss
Perf: the current fairseq has similar speed as vanilla version. After the enablement, the speedup is similar to standalone BT test.
With batch size=64
For V100, the speedup reaches to 1.23x
For A100, the speedup reaches to 1.38x

After enable nested tensor,
For V100, the speedup reaches to 2.46x

Reviewed By: mikekgfb

Differential Revision: D37082681

### DIFF
--- a/fairseq/modules/transformer_layer.py
+++ b/fairseq/modules/transformer_layer.py
@@ -7,14 +7,13 @@ from typing import Dict, List, Optional
 
 import torch
 import torch.nn as nn
+from torch import Tensor
+
 from fairseq import utils
+from fairseq.models.transformer import TransformerConfig
 from fairseq.modules import LayerNorm, MultiheadAttention
 from fairseq.modules.fairseq_dropout import FairseqDropout
 from fairseq.modules.quant_noise import quant_noise
-from torch import Tensor
-from fairseq.models.transformer import (
-    TransformerConfig,
-)
 
 
 class TransformerEncoderLayerBase(nn.Module):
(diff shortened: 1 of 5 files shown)


mikekgfb (approved):

fairseq/modules/transformer_layer.py
```diff
@@ -187,44 +283,83 @@ def forward(
         # Note that we cannot use -inf here, because at some edge cases,
         # the attention weight (before softmax) for some padded element in query
         # will become -inf, which results in NaN in model parameters
-        if attn_mask is not None:
-            attn_mask = attn_mask.masked_fill(
-                attn_mask.to(torch.bool), -1e8 if x.dtype == torch.float32 else -1e4
+
+        if self.training:
+            self.ever_training = True
```
dianaml0, line 288: Why have a variable that is the same as self.training?

tests/test_sequence_generator.py
```diff
@@ -113,7 +140,9 @@ def _test_save_and_load(self, scripted_module):
 JIT_MSG = "Targeting OSS scriptability for the 1.6 release"
 
 
-@unittest.skipIf(torch.__version__ < "1.6.0", JIT_MSG)
+@unittest.skipIf(
```
dianaml0, line 143: You can add the necessary pytorch version to the CI testing to make sure this test works and passes.

fairseq/modules/transformer_layer.py
frank-wei: That is for the scenario where loading ckpt+continue training. We'd like to prevent this case to go fast path.

tests/test_sequence_generator.py
frank-wei: I can add one nightly version like 1.13.0.dev20220613 as one more test.

dianaml0: That sounds good to me

frank-wei:
@dianaml0  The flow creation could be separate from this PR so I landed it first. 
I am not allowed to create any flow or yaml file as I met such error:
 new_flow -> new_flow (refusing to allow a Personal Access Token to create or update workflow `.github/workflows/build_1_13.yml` without `workflow` scope)
 
 Could you help to create a flow with new yaml file? It should be 99% the same with your current `build.yml` but changed the line 31 to:
```
 run: pip3 install --pre torch==1.13.0.dev20220613 -f https://download.pytorch.org/whl/nightly/torch_nightly.html 
```

PR #4593, opened 2022-07-20
Support direct S2ST models in HF demo

sravyapopuri388:
Update the existing S2T and TTS hub interfaces to support S2UT model plus vocoder in the HF demo. As next steps will test and update the HF pipeline. The detailed changes in this PR are

1. Update CodeHifiGAN vocoder to inherit BaseFairseqModel instead of nn.Module to use the underlying hub interface. Also add relevant methods to the vocoder to load and infer using the models from S3
2. Add VocoderHubInterface class to TTS hub interface to support waveform generation using CodeHifiGAN vocoders.
3. Add S2UT models to xm_transformer and update S2T interface accordingly
4. Tested the changes in a script in my local using the script located at
/private/home/spopuri/speech_translation/fairseq_speech/test_HF_demo.py. Couldn't use google collab for testing due to GPU memory constraints

### DIFF
--- a/fairseq/models/text_to_speech/hub_interface.py
+++ b/fairseq/models/text_to_speech/hub_interface.py
@@ -137,3 +137,49 @@ class TTSHubInterface(nn.Module):
     ) -> Tuple[torch.Tensor, int]:
         sample = self.get_model_input(self.task, text, speaker, verbose=verbose)
         return self.get_prediction(self.task, self.model, self.generator, sample)
+
+
+class VocoderHubInterface(nn.Module):
+    def __init__(self, cfg, model):
+        super().__init__()
+        self.vocoder = model
+        self.vocoder.eval()
+        self.sr = 16000
+        self.multispkr = self.vocoder.model.multispkr
+        if self.multispkr:
+            logger.info("multi-speaker vocoder")
+            self.num_speakers = cfg.get(
+                "num_speakers",
+                200,
+            )  # following the default in codehifigan to set to 200
+
+    def get_model_inout(
+        self,
+        text: str,
+        speaker: Optional[int] = None,
+    ):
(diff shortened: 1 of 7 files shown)


jmp84 (approved):

fairseq/models/text_to_speech/hub_interface.py
```diff
@@ -137,3 +137,49 @@ def predict(
     ) -> Tuple[torch.Tensor, int]:
         sample = self.get_model_input(self.task, text, speaker, verbose=verbose)
         return self.get_prediction(self.task, self.model, self.generator, sample)
+
+
+class VocoderHubInterface(nn.Module):
```
kahne, line 142: Can you add a comment (or rename) mentioning that this is currently for unit vocoder only (there are differences in spectrogram vocoder, for example, we normally use 22khz sample rate)?

fairseq/models/text_to_speech/hub_interface.py
```diff
@@ -137,3 +137,49 @@ def predict(
...
+        units = list(map(int, text.strip().split()))
+        x = {
+            "code": torch.LongTensor(units).view(1, -1),
+        }
+
+        if self.multispkr:
+            assert (
+                speaker < self.num_speakers
+            ), f"invalid --speaker-id ({speaker}) with total #speakers = {self.num_speakers}"
+            spk = random.randint(0, self.num_speakers - 1) if speaker == -1 else speaker
```
kahne, line 171: Do you mean `if speaker is None`?

kahne (approved):

fairseq/models/text_to_speech/hub_interface.py
sravyapopuri388: Good catch. Changed the default value to -1

PR #4594, opened 2022-07-20
Merge source and target embedding tables into one table

rasto2211:
Add support for merging source and target embedding table into one table. This feature might hurt performance, but it will decrease the size of the final model.


dianaml0 (approved):
LGTM, perhaps you can add a comment about the tradeoffs in the code as well

dianaml0 (changes requested):
Actually looks like there's an error: 

`AttributeError: invalid argument merge_src_tgt_embed.`

dianaml0 (approved):

PR #4599, opened 2022-07-21
Minor update to end-to-end NLU config

padentomasello:


arbabu123 (approved):
please add world_size too.

PR #4613, opened 2022-07-27
padding fix in the adaptor layer

uralik:
The input to adaptor layer was not padded properly so that resulting downsampled representation contained noise influenced by different batch sizes (i.e., depending on the batch max length). This fix ensures proper zeroing of padding positions of the encoder output before and after applying the convolutional downsamplings. 

The latter zeroing operation is a bit redundant, but it ensures that in potential extra processing the padded information won't influence results.

We verified internally that this fix alleviates the issue of score inconsistency w.r.t. to different values of batch size.


sravyapopuri388 (approved):
Hi Ilia, thanks for digging deep into this. The changes look good to me!

PR #4662, opened 2022-08-22
Fix under generation issue for speech to speech translation models by adding optional generation args

sravyapopuri388:


fairseq/hub_utils.py
```diff
@@ -83,6 +83,9 @@ def from_pretrained(
             model_path,
             arg_overrides=kwargs,
         )
+    if "generation_args" in kwargs:
+        for key in kwargs["generation_args"]:
```
kahne, line 87: Does this cover the case of `generation_args=None`?

sravyapopuri388: Good catch! Fixed it.

kahne (approved):

PR #4670, opened 2022-08-26
UnitY implementation

hirofumi0810:
# Before submitting

## What does this PR do?
This PR supports new speech-to-speech translation models based on two-pass decoders.
- UnitY (text->unit)
- Translatotron2 (text->spectrogram)

These are related to this PR
- https://github.com/facebookresearch/fairseq/pull/4707
- https://github.com/facebookresearch/fairseq/pull/4721
- https://github.com/facebookresearch/fairseq/pull/4722
- https://github.com/facebookresearch/fairseq/pull/4723
- https://github.com/facebookresearch/fairseq/pull/4724
- https://github.com/facebookresearch/fairseq/pull/4730
- https://github.com/facebookresearch/fairseq/pull/4732
- https://github.com/facebookresearch/fairseq/pull/4747
- https://github.com/facebookresearch/fairseq/pull/4763
- https://github.com/facebookresearch/fairseq/pull/4764

In addition to supporting new models, I also made several updates that may affect other tasks, 
- Support dual cross-attention for Transformer decoder

Each of them is necessary to obtain the best result with UnitY.

## PR review

## Did you have fun?


fairseq/criterions/label_smoothed_cross_entropy.py
```diff
@@ -156,6 +183,12 @@ def reduce_metrics(cls, logging_outputs) -> None:
                 if meters["total"].sum > 0
                 else float("nan"),
             )
+        rdrop_kl_loss = utils.item(
+            sum(log.get("rdrop_kl_loss", 0) for log in logging_outputs)
+            # sum(log.get("rdrop_kl_loss", 0) for log in logging_outputs) / sample_size / math.log(2)
```
sravyapopuri388: nit: remove comment as is it is not needed

fairseq/criterions/speech_to_speech_criterion.py
```diff
@@ -308,3 +427,112 @@ def reduce_metrics(cls, logging_outputs) -> None:
             return
 
         MultitaskCriterion.reduce_metrics(logging_outputs)
+
+
+@register_criterion(
```
sravyapopuri388: nit: Is this change related to the UnitY model?

fairseq/data/audio/speech_to_speech_dataset.py
```diff
@@ -231,57 +232,6 @@ def collater(
         return out
 
 
-class TextTargetMultitaskData(object):
```
sravyapopuri388: IUC this will be used for multitasking in vanilla S2UT model https://github.com/facebookresearch/fairseq/blob/main/examples/speech_to_speech/docs/direct_s2st_discrete_units.md The implementation replicated in speech_to_text_dataset seems to be slightly different.
 
 nit: Verify that the vanilla S2UT with multitask training works as expected

fairseq/data/audio/speech_to_text_dataset.py
```diff
@@ -348,25 +509,39 @@ def _from_list(
...
+        dataset_cls = (
+            SpeechToTextMultitaskDataset if has_multitask else SpeechToTextDataset
+        )
+
+        ds = dataset_cls(
             split_name,
             is_train_split,
             cfg,
             audio_paths,
             n_frames,
```
sravyapopuri388: nit: keep the argument names intact as it helps with code readibility

fairseq/models/speech_to_text/xm_transformer.py
```diff
@@ -623,15 +640,10 @@ def build_model(cls, args, task):
...
-            args = cls.override_decoder_args(args, decoder_args_dict)
-
-        def build_embedding(dictionary, embed_dim):
-            num_embeddings = len(dictionary)
-            padding_idx = dictionary.pad()
-            return Embedding(num_embeddings, embed_dim, padding_idx)
+        # if getattr(args, "load_pretrained_decoder_from", None) is not None:
+        #     ckpt = torch.load(getattr(args, "load_pretrained_decoder_from", None))
+        #     decoder_args_dict = cls.get_decoder_args_from_checkpoint(ckpt["cfg"])
+        #     args = cls.override_decoder_args(args, decoder_args_dict)
```
uralik: This one is commented now, because currently we want to override some of the pretrained mbart parameters: we remove learned pos emb and layernorm from the embedding of that model. However, turning this off completely may lead to unintentional problems later when we forgot to match all the mbart parameters from pretrained model which are different from the default. 
I wonder if we could utilize model-override arg here somehow to make sure that we can do both loading the cfg from the pretrained mbart, and then add / change it with args from the CLI.

fairseq/criterions/label_smoothed_cross_entropy.py
```diff
@@ -59,6 +64,7 @@ def __init__(
         task,
         sentence_avg,
         label_smoothing,
+        rdrop_alpha,
```
sravyapopuri388: Add a default value for backward compatibility of existing models

fairseq/criterions/label_smoothed_cross_entropy.py
```diff
@@ -112,7 +131,15 @@ def compute_loss(self, model, net_output, sample, reduce=True):
...
-        return loss, nll_loss
+
+        if self.rdrop_alpha > 0:
+            pad_mask = target[: target.size(0) // 2].unsqueeze(-1).eq(self.padding_idx)
+            rdrop_kl_loss = compute_kl_loss(model, net_output, pad_mask)
+            loss += self.rdrop_alpha * rdrop_kl_loss
+        else:
+            rdrop_kl_loss = loss.new_zeros(1)
+
+        return loss, nll_loss, rdrop_kl_loss
```
sravyapopuri388: Don't change the default list of objects being returned to ensure backward compatibility.

fairseq/data/audio/speech_to_speech_dataset.py
```diff
@@ -297,8 +247,9 @@ def __getitem__(
 
         multitask_target = {}
         sample_id = self.ids[index]
+        tgt_lang = self.tgt_langs[index]
         for task_name, task_dataset in self.multitask_data.items():
-            multitask_target[task_name] = task_dataset.get(sample_id)
+            multitask_target[task_name] = task_dataset.get(sample_id, tgt_lang)
```
sravyapopuri388, line 252: IRC this will be used for multitasking in vanilla S2UT model https://github.com/facebookresearch/fairseq/blob/main/examples/speech_to_speech/docs/direct_s2st_discrete_units.md and that model currently doesn't use language tag. Do you think this change is still compatible with the existing models?

fairseq/models/speech_to_text/xm_transformer.py
sravyapopuri388: Agree with Ilia that commenting it out might not be a good idea. Can we throw a warning here instead of commenting it out till we figure out a long term solution?

fairseq/models/speech_to_text/xm_transformer.py
```diff
@@ -641,6 +653,37 @@ def build_embedding(dictionary, embed_dim):
...
+                dictionary=tgt_dict,
+                in_dim=in_dim,
+            )
+        else:
+            raise NotImplementedError(
+                "currently only support multitask decoder_type 'transformer', 'ctc'"
+            )
+
+        return task_decoder
+
```
sravyapopuri388: nit: Move this to xm_transformer_unity since this is not being used here

fairseq/models/transformer/transformer_decoder_aug.py
```diff
@@ -0,0 +1,391 @@
...
+from fairseq.models.transformer.transformer_decoder import TransformerDecoderBase
+from fairseq.modules import (
+    LayerDropModuleList,
+    SinusoidalPositionalEmbedding,
+    transformer_layer_aug,
+)
+from fairseq.modules.checkpoint_activations import checkpoint_wrapper
+
+
+class AugTransformerDecoderBase(TransformerDecoderBase):
```
sravyapopuri388, line 24: nit: Add description on how this is different from exiting transformer implementation

fairseq/sequence_generator.py
```diff
@@ -245,18 +237,55 @@ def _generate(
...
-        beam_size = self.beam_size
-
         if constraints is not None and not self.search.supports_constraints:
             raise NotImplementedError(
                 "Target-side constraints were provided, but search method doesn't support them"
             )
 
         # Initialize constraints, when active
-        self.search.init_constraints(constraints, beam_size)
+        self.search.init_constraints(constraints, self.beam_size)
```
sravyapopuri388: nit: Verify that the changes work as expected for exiting models to ensure backward compatibility.

fairseq/speech_generator.py
```diff
@@ -76,7 +76,203 @@ def generate(self, model, sample, has_targ=False, **kwargs):
                 incremental_state=incremental_state,
                 target_lengths=cur_out_lens,
                 speaker=sample["speaker"],
-                **kwargs
+                **kwargs,
+            )
+            cur_eos_prob = torch.sigmoid(cur_eos_out).squeeze(2)
```
sravyapopuri388, line 81: Do we know if this change is backward compatible with the existing models

fairseq/tasks/speech_to_speech.py
```diff
@@ -265,15 +283,15 @@ def setup_task(cls, args, **kwargs):
...
-        if len(self.multitask_tasks) > 0:
-            if self.args.target_is_code and args._name != "speech_to_unit":
-                raise ValueError(
-                    "set --criterion speech_to_unit for speech-to-unit loss with multitask"
-                )
-            elif not self.args.target_is_code and args._name != "speech_to_spectrogram":
-                raise ValueError(
-                    "set --criterion speech_to_spectrogram for speech-to-spectrogram loss with multitask"
-                )
+        # if len(self.multitask_tasks) > 0:
```
sravyapopuri388: Curious, why is this commented? Do we know if vanilla S2UT models with multitasking work as expected when this is commented out?

sravyapopuri388 (changes requested):
Overall the changes look decent. Left inline comments but listing few high level comments here
1. Make sure all the changes are backward compatible with exiting models and don't affect the existing models. Suggest not changing any of the existing code at all if possible. If not suggest retraining models whose code is changed to make sure the performance is as expected. For ex: vanilla S2UT model with pretraining, evaluation of existing S2T/S2ST models (since tasks, sequence generator are being modified)
2. Could we ensure that all the build tests here pass https://github.com/facebookresearch/fairseq/runs/8050111393?check_suite_focus=true Also run all unit tests on devfair (some of them rely on gpu) https://github.com/facebookresearch/fairseq/tree/main/tests to ensure nothing breaks
3. Check with Diana to make sure updating sequence generator is ok

fairseq/criterions/label_smoothed_cross_entropy.py
```diff
@@ -165,3 +199,44 @@ def logging_outputs_can_be_summed() -> bool:
         to True will improves distributed training speed.
         """
         return True
+
+
+def duplicate_input(sample):
```
an918tw: nit: this function directly related to label_smoothed_cross_entropy, so we should consider moving it to some utils files

fairseq/criterions/speech_to_speech_criterion.py
```diff
@@ -208,6 +251,82 @@ def logging_outputs_can_be_summed() -> bool:
...
+)
+class SpeechToUnitTranslatotron2MultitaskTaskCriterion(
+    SpeechToUnitMultitaskTaskCriterion
+):
+    def __init__(
+        self,
+        task,
+        sentence_avg,
+        label_smoothing,
+        rdrop_alpha,
```
an918tw: is this required, or should there be a default value?

fairseq/criterions/speech_to_speech_criterion.py
```diff
@@ -308,3 +427,110 @@ def reduce_metrics(cls, logging_outputs) -> None:
             return
 
         MultitaskCriterion.reduce_metrics(logging_outputs)
+
+
+@register_criterion("speech_to_spectrogram_2pass", dataclass=Tacotron2CriterionConfig)
+class SpeechToSpectrogramTranslatotron2MultitaskTaskCriterion(
+    Tacotron2Criterion, MultitaskCriterion
```
an918tw: why not `SpeechToSpectrogramMultitaskTaskCriterion`? We don't need to copy `reduce_metrics` for example.

fairseq/data/audio/text_to_speech_dataset.py
```diff
@@ -196,6 +197,7 @@ def _from_list(
         bpe_tokenizer,
         n_frames_per_step,
         speaker_to_id,
+        multitask=None,
```
aghatare: Doesn't seem to be used anywhere else, what is this for?

fairseq/models/speech_to_speech/s2s_conformer.py
```diff
@@ -57,31 +53,68 @@ def forward(
...
 
     @classmethod
     def build_encoder(cls, args):
-        print(args)
-        data_cfg = S2SDataConfig(Path(args.data) / args.config_yaml)
-        args.input_feat_per_channel = data_cfg.input_feat_per_channel
-        args.input_channels = data_cfg.input_transformed_channels
+        encoder = S2SConformerEncoder(args)
+        pretraining_path = getattr(args, "load_pretrained_encoder_from", None)
+        if pretraining_path is not None:
```
aghatare: (Just curious) Is the code duplication between TranslatotronConformer and S2UTConformer intentional/necessary?

hirofumi0810: @aghatare Thank you for your suggestion. I'll unify these two methods.

fairseq/data/audio/text_to_speech_dataset.py
hirofumi0810: @aghatare This is necessary because TextToSpeechDataset inherits SpeechToTextDataset.

fairseq/criterions/speech_to_speech_criterion.py
hirofumi0810: @sravyapopuri388 This is for Translatotron2.

fairseq/data/audio/speech_to_speech_dataset.py
an918tw: +1 on making sure backward compatibility

fairseq/models/speech_to_speech/s2s_transformer.py
```diff
@@ -414,8 +416,8 @@ def forward(
         return decoder_out
 
 
-@register_model("s2spect_transformer")
```
an918tw: Be mindful about backward compatibility. We chose to name the model with s2spec on purpose, since it's clearer than what "Translatotron" means, and our implementation is not exactly the same as the original Translatotron model, which was LSTM-based anyway.

fairseq/models/speech_to_speech/s2s_conformer_translatotron2.py
```diff
@@ -0,0 +1,261 @@
...
+                else args.decoder_embed_dim
+            )
+            task_decoder = cls.build_multitask_decoder(
+                task_obj.args,
+                task_obj.target_dictionary,
+                in_dim,
+                is_mt_decoder,
+                getattr(args, "translation_decoder_layers", 4),
+                getattr(args, "decoder_embed_dim", 256),
+                getattr(args, "decoder_attention_heads", 4),
```
an918tw, line 144: curious why these default values are defined here but not part of register_model_architecture below?

fairseq/models/speech_to_speech/s2s_transformer.py
```diff
@@ -664,9 +675,9 @@ def s2ut_architecture_fisher(args):
 
 
 @register_model_architecture(
-    model_name="s2spect_transformer", arch_name="s2spect_transformer"
+    model_name="translatotron_transformer", arch_name="translatotron_transformer"
```
an918tw: please test loading old models

fairseq/tasks/speech_to_speech.py
```diff
@@ -208,15 +213,28 @@ def __init__(self, args, tgt_dict, infer_tgt_lang_id=None):
...
+                self.multitask_tasks[task_name] = task_obj
+                if "target" in task_name and task_obj.args.decoder_type != "ctc":
+                    self.tgt_dict_mt = task_obj.target_dictionary
+                    if task_config.prepend_bos_and_append_tgt_lang_tag:
+                        self.eos_token_mt = task_config.eos_token
+                        assert not isinstance(self.eos_token_mt, List)
+
+                        if not self.eos_token_mt:
+                            raise Warning(
+                                "Please provide --eos_token to replace eos in sequence generator"
```
an918tw: nit: not really `--eos_token` but more like `eos_token` in the config file?

fairseq/tasks/speech_to_speech.py
```diff
@@ -208,15 +213,28 @@ def __init__(self, args, tgt_dict, infer_tgt_lang_id=None):
...
             multitask_cfg = MultitaskConfig(
                 Path(args.data) / args.multitask_config_yaml
             )
             for task_name, task_config in multitask_cfg.get_all_tasks().items():
-                self.multitask_tasks[task_name] = DummyMultiTask(
-                    task_config, task_config.tgt_dict
-                )
+                task_obj = DummyMultiTask(task_config, task_config.tgt_dict)
+                self.multitask_tasks[task_name] = task_obj
+                if "target" in task_name and task_obj.args.decoder_type != "ctc":
```
an918tw: Is this mainly for the first decoder target for dual decoder architecture? I wonder if we shouldn't solely rely on `"target" in task_name` but add other fields to mark this special type of multitask, since previously we can also have an encoder multitask named "target_letter", "target_phoneme", or anything we'd like. Having an additional indicator for first decoder target will also allow us not to rely on the assumption that the last task is for the first-pass decoder.

fairseq/tasks/speech_to_speech.py
an918tw: I'm guessing it's because we have speech_to_unit_2pass and speech_to_spectrogram_2pass, two new criterion now. We can just check the prefix of `self.__name`.

fairseq/tasks/speech_to_speech.py
```diff
@@ -343,14 +395,26 @@ def build_generator(
...
+
         if self.args.target_is_code:
             if self.args.n_frames_per_step == 1:
-                seq_generator = super().build_generator(
-                    models,
-                    args,
-                    seq_gen_cls=None,
-                    extra_gen_cls_kwargs=extra_gen_cls_kwargs,
-                )
+                if isinstance(models[0], UnitYConformerModel):
```
an918tw: Does only `UnitYConformerModel` need `MultiDecoderSequenceGenerator`? How about other models with dual-decoder style architectures? Maybe it's easier to have something like `self.has_dual_decoder` in this task, which we can set the value when we read and parse `args.multitask_config_yaml`.

an918tw (changes requested):
This is a HUGE PR! +1 to what Sravya suggested. Please
1. split the PR into smaller ones. Put code refactoring/formatting in separate PR. Also put changes that affect general code like sequence_generator in separate PR. 
2. Run all tests and check backward compatibility, both in terms of model training and model loading. For previous checkpoints (S2UT models w/ or w/o multitasks, Sravya should have a few) that cannot be loaded due to code change, we need to add mitigation, e.g. upgrade_state_dict or something similar.

Looking forward to releasing the code!

fairseq/criterions/ctc.py
```diff
@@ -106,12 +111,31 @@ def __init__(self, cfg: CtcCriterionConfig, task: FairseqTask):
...
             net_output, log_probs=True
         ).contiguous()  # (T, B, C) from the encoder
 
+        # CTC loss is calculated over duplicated inputs
+        # sample is already duplicated for R-Drop
+        if self.rdrop_alpha > 0:
+            for k, v in sample.items():
+                if k in ["target", "target_lengths"]:
+                    sample[k] = torch.cat([v, v.clone()], dim=0)
+                elif k == "net_input":
```
yuntang: Some datasets use "source" or "features" instead of "src_tokens".
Is the rdrop loss calculated with the CTC class? I don't see it here.

fairseq/criterions/label_smoothed_cross_entropy.py
```diff
@@ -61,23 +66,32 @@ def __init__(
...
 
         Returns a tuple with three elements:
         1) the loss
         2) the sample size, which is used as the denominator for the gradient
         3) logging outputs to display while training
         """
-        net_output = model(**sample["net_input"])
-        loss, nll_loss = self.compute_loss(model, net_output, sample, reduce=reduce)
+        if net_output is None:
+            if self.rdrop_alpha > 0 and sample["net_input"]["src_tokens"].size(
```
yuntang: We might need to consider "source" or "features" too.

fairseq/criterions/speech_to_speech_criterion.py
hirofumi0810: @an918tw This is required because `SpeechToUnitTranslatotron2MultitaskTaskCriterion` inherits `SpeechToUnitMultitaskTaskCriterion`.

fairseq/criterions/speech_to_speech_criterion.py
hirofumi0810: @an918tw This is because `prev_output_tokens_mt` is needed in `SpeechToSpectrogramTranslatotron2MultitaskTaskCriterion`.

fairseq/models/speech_to_speech/modules/stacked_embedding.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
yuntang: Why not move this module to fairseq/modules/ directory?
Can you give description for the main purpose of this embedding?

fairseq/models/speech_to_text/xm_transformer_unity.py
```diff
@@ -0,0 +1,345 @@
...
+        encoder = cls.build_encoder(args)
+        decoder, proj, unit_args = cls.build_decoder(
+            args,
+            task,
+            aug_attn=getattr(args, "synthesizer_augmented_cross_attention", False),
+        )
+        base_model = cls(encoder, decoder)
+        setattr(base_model, "proj", proj)
+
+        base_model.t2u_augmented_cross_attn = getattr(
```
yuntang, line 177: Can we do this within the \_\_init\_\_ function? decoder should include this information.

fairseq/models/speech_to_text/xm_transformer_unity.py
```diff
@@ -0,0 +1,345 @@
...
+        base_model.multitask_decoders = {}
+        n_aux_tasks = len(list(task.multitask_tasks.items()))
+        for i, (task_name, task_obj) in enumerate(task.multitask_tasks.items()):
+
+            if i < n_aux_tasks - 1:
+                task_decoder = cls.build_multitask_decoder(
+                    task_obj.args, task_obj.target_dictionary, args.decoder_embed_dim
+                )
+            else:
+                base_model.mt_task_name = task_name
```
yuntang, line 188: Do we really need to set the mt_decoder dynamically in the configure file? I mean if we could fix this in class definition, the code will be much easier to read.

fairseq/models/transformer/transformer_decoder_aug.py
yuntang: +1

fairseq/modules/transformer_layer_aug.py
```diff
@@ -0,0 +1,319 @@
...
+            cfg,
+            no_encoder_attn=no_encoder_attn,
+            add_bias_kv=add_bias_kv,
+            add_zero_attn=False,
+        )
+        assert not no_encoder_attn
+
+        self.encoder_attn = self.build_encoder_attention(self.embed_dim, cfg)
+        self.encoder_attn_layer_norm = LayerNorm(self.embed_dim, export=cfg.export)
+        self.encoder_attn2 = self.build_encoder_attention(self.embed_dim, cfg)
```
yuntang, line 48: So attn2 will be used to attend another encoder outputs, say speech encoder output, is that correct?

fairseq/criterions/label_smoothed_cross_entropy.py
hirofumi0810: @yuntang When do we use "source" or "features"?

yuntang: The outputs from HubertDataset and RawAudioDataset are "source" instead of "src_tokens". 
Also, we can find "source", "src_tokens" and "features" are all legal input types in sequence_generator
https://github.com/facebookresearch/fairseq/blob/main/fairseq/sequence_generator.py#L222-L241
So I guess we might meet "source" and "features" in the cost function too.

fairseq/models/speech_to_speech/modules/stacked_embedding.py
hirofumi0810: @yuntang This is used for S2UT models only.

fairseq/modules/transformer_layer_aug.py
hirofumi0810: @yuntang Correct.

fairseq/models/speech_to_text/xm_transformer_unity.py
hirofumi0810: @yuntang I think this method (`build_model`) is called before the initialization of the class.

fairseq/models/speech_to_text/xm_transformer_unity.py
hirofumi0810: @yuntang Ditto.

fairseq/criterions/label_smoothed_cross_entropy.py
hirofumi0810: @yuntang, I think both HubertDataset and RawAudioDataset are used during pre-training only. So we never use "source" or "features" during fine-tuning.

fairseq/models/speech_to_text/xm_transformer_unity.py
yuntang: base_model is created at line 171.

fairseq/models/speech_to_text/xm_transformer_unity.py
yuntang: Not sure I get this. We can keep the best configure and hard code it in the class definition instead of using a configure file to set it up. It will make the code much easier to read. However, if you feel other configures might be useful, we can keep the current form.

fairseq/models/speech_to_text/xm_transformer.py
hirofumi0810: @uralik @sravyapopuri388 Thank you for your feedback. I reverted the change because it doesn't affect two-pass models now.

fairseq/tasks/speech_to_speech.py
hirofumi0810: @an918tw Thank you for the useful feedback! I added a flag of "is_first_pass_decoder" in the multitask config file. If it is not specified, the last non-CTC decoder will be used as a first-pass decoder.

examples/speech_to_speech/unity/sequence_generator.py
```diff
@@ -0,0 +1,1068 @@
+# Copyright (c) Facebook, Inc. and its affiliates.
```
sravyapopuri388, line 1: I see that most of this code is borrowed from sequence_generator.py Wondering if there is a way to use the existing implementation without copying it over

fairseq/data/audio/speech_to_speech_dataset.py
sravyapopuri388: Would be great if we can make sure the existing vanila S2UT models run on this PR as expected

fairseq/tasks/speech_to_speech.py
```diff
@@ -209,15 +213,33 @@ def __init__(self, args, tgt_dict, infer_tgt_lang_id=None):
...
+        self.tgt_dict_mt = None
+        self.eos_token_mt = None
         if getattr(args, "multitask_config_yaml", None) is not None:
             multitask_cfg = MultitaskConfig(
                 Path(args.data) / args.multitask_config_yaml
             )
-            for task_name, task_config in multitask_cfg.get_all_tasks().items():
-                self.multitask_tasks[task_name] = DummyMultiTask(
-                    task_config, task_config.tgt_dict
+            first_pass_task_idx = multitask_cfg.first_pass_decoder_task_index
```
sravyapopuri388, line 224: This might throw an error for vanila S2UT model as this parameter might not be defined there.

sravyapopuri388 (commented):
The code changes look decent to me but I am slightly worried about the backward compatibility of the changes with the existing vanila S2UT models. Would highly encourage you to double check that the training and inference for vanila S2UT models work as expected.

Here are some models in case you want to verify : /checkpoint/spopuri/H2_2021/discrete_S2ST/S2U/best_ckpts/S2U_encoder_MT/checkpoint_best.pt (es-en with src letter and tgt letter multitasks on the encoder)

examples/speech_to_speech/unity/sequence_generator.py
hirofumi0810: @sravyapopuri388 I copied it because it was difficult to reuse the current sequence generation without significant change.

hirofumi0810: @sravyapopuri388 I inherited SequenceGenerator.

fairseq/tasks/speech_to_speech.py
hirofumi0810: @sravyapopuri388 We use this line only when multitask_config_yaml is set. So there should be no error with S2UT+multitask.

fairseq/criterions/speech_to_speech_criterion.py
hirofumi0810: @an918tw You are right. I fixed this.

fairseq/sequence_generator.py
hirofumi0810: @sravyapopuri388 Close because of `examples/unity`.

yuntang (approved):

fairseq/models/speech_to_text/xm_transformer_unity.py
hirofumi0810: @yuntang I think it is difficult to do so in __init__ because of the design of fairseq.

PR #4673, opened 2022-08-29
[nllb] Release NLLB-200 translations

vedanuj:
Release NLLB-200 MoE 54.5B model translations.

Co-authored-by: Maha Elbayad <elbayadm@users.noreply.github.com>


fiqas:
What is the status of NLLB MoE data being released?

PR #4685, opened 2022-08-31
Update deprecated torch.qr in glow.py example

kit1980:
torch.qr is deprecated for a long time and is being removed by https://github.com/pytorch/pytorch/pull/70989.

This PR makes the example compatible with new and old PyTorch versions.


kit1980:
The test errors seem unrelated - "OSError: Could not connect to socket".

PR #4847, opened 2022-11-03
Fix Torchscript typing in transformer_encoder.py

zhxchen17:
# Before submitting

- [ ] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/main/CONTRIBUTING.md)?

## What does this PR do?
Fixes # (issue).

## PR review

## Did you have fun?

### DIFF
--- a/fairseq/models/transformer/transformer_encoder.py
+++ b/fairseq/models/transformer/transformer_encoder.py
@@ -202,9 +202,12 @@ class TransformerEncoderBase(FairseqEncoder):
         """
         # compute padding mask
         encoder_padding_mask = src_tokens.eq(self.padding_idx)
-        has_pads: Tensor = (
+        has_pads = (
             torch.tensor(src_tokens.device.type == "xla") or encoder_padding_mask.any()
         )
+        # Torchscript doesn't handle bool Tensor correctly, so we need to work around.
+        if torch.jit.is_scripting():
+            has_pads = torch.tensor(1) if has_pads else torch.tensor(0)
 
         x, encoder_embedding = self.forward_embedding(src_tokens, token_embeddings)
 


zhxchen17:
@alexeib 

I think this should fix the Torchscript compilation error introduced in https://github.com/facebookresearch/fairseq/pull/4818

meanwhile is there recommended test to run in addition to the CI tests?

alexeib:
some tests are still failing

zhxchen17:
> some tests are still failing

@alexeib ok after some investigation, seems these build failures are introduced by PyTorch 1.13.0 version release. (which is on Oct 28)

After I install pytorch version 1.12.1 locally and rerun the tests by `pytest`, seems the errors from
`tests/test_binaries.py::TestTranslation::test_transformer_pointer_generator` and `tests/test_multihead_attention.py::test_xformers_single_forward_parity` are gone.

failures under tests/test_plasma_utils.py::TestPlasmaView seems irrelevant:
```
OSError: Could not connect to socket /tmp/tmphatl4hub
```

I guess we could proceed with this PR and fix other issues separately?

fairseq/models/transformer/transformer_encoder.py
```diff
@@ -202,9 +202,12 @@ def forward_scriptable(
         """
         # compute padding mask
         encoder_padding_mask = src_tokens.eq(self.padding_idx)
-        has_pads: Tensor = (
+        has_pads = (
             torch.tensor(src_tokens.device.type == "xla") or encoder_padding_mask.any()
         )
+        # Torchscript doesn't handle bool Tensor correctly, so we need to work around.
+        if torch.jit.is_scripting():
```
tugsbayasgalan, line 209: Can you also add torch.jit.is_tracing()?

tugsbayasgalan (approved):

fairseq/models/transformer/transformer_encoder.py
zhxchen17: @tugsbayasgalan after second thought I don't think we need to handle jit.tracing case. If you take a look at this line in the same function:
```
if return_all_hiddens and not torch.jit.is_scripting():
```

it doesn't account for tracing case as well. proactively adding a tracing guard here might not have any real use case.

alexeib:
thanks for investigating @zhxchen17 . Shall we open another issue for the pytorch test breakage  and reference it here?

zhxchen17:
> thanks for investigating @zhxchen17 . Shall we open another issue for the pytorch test breakage and reference it here?

sure lemme open an issue.

PR #4848, opened 2022-11-03
Add gpu support for speech-to-text for input processing

artkorenev:
Before this PR `get_model_input` returns CPU tensors for input tokens which prevents working with GPU models. I added logic similar to [speech_to_speech/hub_interface.py](https://github.com/artkorenev/fairseq/blob/ust/fairseq/models/speech_to_speech/hub_interface.py#L67) where the input is put on the device if needed. Similarly I also added putting the model to the device if the device is available.


sravyapopuri388 (approved):
Looks good

Issue #3799, opened 2021-08-18
[Wav2Vec2] Cannot load newly added Wav2Vec2 checkpoints


patrickvonplaten:
## 🐛 Bug

A recent commit: https://github.com/pytorch/fairseq/commit/2513524a1604dbafcc4ea9cc5a99ae0aa4f19694 added two new fine-tuned Wav2Vec2 checkpoints, however it seems like there is a problem with the saved config as one cannot load those checkpoints. E.g. the following code cannot be run:

```python
import fairseq
model, _, _ = fairseq.checkpoint_utils.load_model_ensemble_and_task([checkpoint_path], arg_overrides={"data": "path/to/dict"})
```

### To Reproduce

The following colab reproduces the error (one just has to run all cells): 
https://colab.research.google.com/drive/13hJI4w8pOD33hxOJ_qwKkN9QqdKVH5IM?usp=sharing

Kindly pinging @alexeib here :-)

abodacs:
@patrickvonplaten 
I think this fixed on master branch
I changed the install line to 
```diff
!pip install git+https://github.com/pytorch/fairseq.git@master
```

patrickvonplaten:
Still got the same problem ;-) 

See https://colab.research.google.com/drive/13hJI4w8pOD33hxOJ_qwKkN9QqdKVH5IM?usp=sharing

abodacs:
@patrickvonplaten 
a notebook with a working example
also, I removed 
```diff
arg_overrides
``` 

https://colab.research.google.com/drive/1gPQ1LzAoEbQtRYRGVPz4zg9Klo-ErwjH?usp=sharing

ag027592:
@patrickvonplaten 
Hi, I met the same problem.
Do you have any solution?
Thank you.
I run the code:

`model, _, _ = fairseq.checkpoint_utils.load_model_ensemble_and_task([cp_path])`

I got the error:

`ConfigKeyError: Key 'target_dict' not in 'AudioPretrainingConfig'
	full_key: target_dict
	reference_type=Optional[AudioPretrainingConfig]
	object_type=AudioPretrainingConfig`

effendijohanes:
Bump. got similar error. 

Both `target_dict` and `eval_wer` does not make sense to be in Audio**Pretraining**Config, as pretraining does not use text. Has someone found the solution? 

```
omegaconf.errors.ConfigKeyError: Key 'eval_wer' not in 'AudioPretrainingConfig'
[...]

PR #4905, opened 2022-12-13
remove missing config entries when loading task from checkpoint

alexeib:
fixes not being able to load older tasks
partially addresses #3799

### DIFF
--- a/fairseq/checkpoint_utils.py
+++ b/fairseq/checkpoint_utils.py
@@ -440,10 +440,12 @@ def load_model_ensemble_and_task(
                 )
 
             if task is None:
-                task = tasks.setup_task(cfg.task)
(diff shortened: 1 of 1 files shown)


alexeib:
fixes not being able to load older tasks
partially addresses https://github.com/facebookresearch/fairseq/issues/3799

Mortimerp9:
@alexeib, this seems to break some test: https://github.com/facebookresearch/fairseq/actions/runs/3683275218/jobs/6231680434

the CI in main are not passing anymore.

Do you have any idea what's up?

alexeib:
is that failure reliably failing on this commit and reliably passing on 1 commit behind ?

the idea here is that if we are loading from a checkpoint, we should not crash if task config definition has changed since then

not sure how it would break this test. can also just add a "build_model" variation to the task without from_checkpoint arg

PR #5080, opened 2023-04-16
fix variable naming bug

orena1:
The function `_get_mask_indices_dims` used the wrong variable when checking if a length was already cached.  This will increase performance


orena1:
Just noticed that it is the same as this one - https://github.com/facebookresearch/fairseq/pull/5018#issue-1618732029 not sure why it was closed

PR #5082, opened 2023-04-17
Preserve order when using skip_invalid_size_inputs

AbdallahNasir:
## What does this PR do?

When translating using a model from hub_utils in the production setting, if skip_invalid_size_inputs is set to True, then the output list of translations may be different in length from the input list of strings.

These changes enforce empty responses in case of failure instead of no responses which will lead to mismatching the input sentences from the hyp outputs when using batching.


erip:
This is a massively helpful PR. I routinely have to re-align my source with my generated output which can be very messy. This is a very easy PR to accept.

AbdallahNasir:
Thanks @erip ! Would appreciate if someone from the team can look into it.

PR #3627, opened 2021-06-17
Enabling word-level timestamps for Wav2Vec 2.0

Nithin-Holla:
# Before submitting

- [ ] Was this discussed/approved via a Github issue? (no need for typos, doc improvements)
- [x] Did you read the [contributor guideline](https://github.com/pytorch/fairseq/blob/master/CONTRIBUTING.md)?
- [ ] Did you make sure to update the docs?   
- [ ] Did you write any new necessary tests?  

## What does this PR do?
Fixes #3371.

Currently, the output from Wav2Vec 2.0 decoding does not contain word-level start/end times, which can be useful for certain applications of ASR. Based on the discussion [here](https://github.com/flashlight/flashlight/issues/618), they could be computed based on the output from the Flashlight decoder. For the KenLM decoder, we could first obtain the frame number corresponding to each non-blank token. Next, the timestamp of each character could be computed as `segment_start + frame_no/total_frames * segment_duration`. Finally, the start and end time of each word could be calculated based on the timestamp of the word boundary characters. In order to enable this, the frame number of each non-blank character is returned as a result of KenLM decoding. This is similar to the `timesteps` output from the [ctcdecode](https://github.com/parlance/ctcdecode#outputs-from-the-decode-method) library.

## PR review    
@alexeib

PR #5403, opened 2023-12-17
Enabling word-level timestamps for all W2L Decoders

### LINKED ISSUE #3371: [Question] wav2vec 2.0 timestamp words
Is it possible to extract the Words start time and end time of the transcript? Thank you!

abarcovschi:
# Before submitting

## What does this PR do?
Fixes #3371 and extends #3627 to include the ability to return the frame numbers of all non-blank characters of a hypothesis for all wav2letter decoder classes, not only just for `W2lKenLMDecoder`. A method called `get_symbols()` was also added to the parent class for all the decoders (`W2lDecoder`) so that the non-blank characters of the hypothesis can be returned as a list of natural language characters and not just token ids. This helps in finding the word-boundary tokens later when calculating the word-level timestamp information using the following formula:

timestamp = frame_num * (audio_len / (num_frames * sample_rate))

where:

- frame_num = the timestep of the symbol, as returned in the 'timesteps' field of Wl2Decoder.decode() outputs.
- audio_len = the number of samples in the loaded audio file corresponding to the transcript (if using batched w2v2 acoustic model inference, will be zero padded to the length of the longest loaded audio file in the batch).
- num_frames = the number of frames in the emission matrix returned by the w2v2 acoustic model inference for that audio file (if using batched inference, the number of frames for each audio file will be the same as in this case all loaded audio files are padded to the length of the longest audio file in the batch).
- sample_rate = sample rate of loaded audio files (usually 16000 Hz).

## PR review
@alexeib

### DIFF
--- a/examples/speech_recognition/w2l_decoder.py
+++ b/examples/speech_recognition/w2l_decoder.py
@@ -12,33 +12,33 @@ Flashlight decoders.
 import gc
 import itertools as it
 import os.path as osp
-from typing import List
 import warnings
 from collections import deque, namedtuple
+from typing import List
 
 import numpy as np
 import torch
+from omegaconf import open_dict
+
 from examples.speech_recognition.data.replabels import unpack_replabels
 from fairseq import tasks
-from fairseq.utils import apply_to_sample
-from omegaconf import open_dict
-from fairseq.dataclass.utils import convert_namespace_to_omegaconf
 from fairseq.data.data_utils import post_process
-
+from fairseq.dataclass.utils import convert_namespace_to_omegaconf
+from fairseq.utils import apply_to_sample
 
 try:
-    from flashlight.lib.text.dictionary import create_word_dict, load_words
     from flashlight.lib.sequence.criterion import CpuViterbiPath, get_data_ptr_as_bytes
     from flashlight.lib.text.decoder import (
-        CriterionType,
-        LexiconDecoderOptions,
-        KenLM,
         LM,
+        CriterionType,
+        KenLM,
+        LexiconDecoder,
+        LexiconDecoderOptions,
         LMState,
         SmearingMode,
         Trie,
-        LexiconDecoder,
     )
+    from flashlight.lib.text.dictionary import create_word_dict, load_words
 except:
     warnings.warn(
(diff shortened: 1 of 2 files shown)


examples/speech_recognition/w2l_decoder.py
```diff
@@ -116,10 +159,21 @@ def decode(self, emissions):
...
+
+        for b in range(B):
+            hypos.append(
+                [
+                    {
+                        "tokens": self.get_tokens(viterbi_path[b].tolist()).tolist(), # non-blank token idxs.
+                        "symbols": self.get_symbols(self.get_tokens(viterbi_path[b].tolist()).tolist()), # characters (symbols) corresponding to non-blank token idxs.
+                        "score": 0,
+                        "timesteps": self.get_timesteps(viterbi_path[b].tolist()), # frame numbers of non-blank tokens.
+                        "words": post_process(self.tgt_dict.string(self.get_tokens(viterbi_path[b].tolist()).int().cpu()), 'letter').split(' ') # the transcript as a list of words.
```
alexeib: maybe call get_tokens once?

examples/speech_recognition/w2l_decoder.py
```diff
@@ -227,12 +260,11 @@ def decode(self, emissions):
...
-                        "tokens": self.get_tokens(result.tokens),
+                        "tokens": self.get_tokens(result.tokens).tolist(), # non-blank token idxs.
+                        "symbols": self.get_symbols(self.get_tokens(result.tokens)), # characters (symbols) corresponding to non-blank token idxs.
                         "score": result.score,
-                        "timesteps": self.get_timesteps(result.tokens),
-                        "words": [
-                            self.word_dict.get_entry(x) for x in result.words if x >= 0
-                        ],
+                        "timesteps": self.get_timesteps(result.tokens), # frame numbers of non-blank tokens.
+                        "words": [self.word_dict.get_entry(x) for x in result.words if x >= 0] # the transcript as a list of words. Empty if lexicon-free decoding.
```
alexeib: also here, get_tokens should be called once

alexeib (changes requested):
lgtm bar comments. also i am no longer at meta so can't merge PRs into this repo

alexeib (approved):
thanks! hopefully someone from meta will merge!

